PR #50, opened 2025-10-24
45 only run the unit tests which touch changed code in the CI

### LINKED ISSUE #45: Only run the unit tests which touch changed code in the CI

max-models:

Closes #45 

**Core changes:**

None

**Model-specific changes:**

None

**Documentation changes:**

None


max-models:
It seems like the `-k` flag isn't working well with `--testmon`. When running the unit tests twice it still runs all of them unless I remove the `-k` flag. One option would be to categorize the tests in separate subdirectories instead and use `--ignore=<PATH>` to specify whether to run unit tests or model tests.

max-models:
> > Seems like testmon might be problematic for MPI tests https://github.com/struphy-hub/struphy/actions/runs/18819880612/job/53694176303
> 
> If this problem persists, we can try to shorten the mpi unit tests (using `pytest.mark.mpi`), for example by replacing the long convergence tests by simpler non-parameter-scan tests.

This could be an option, but I think I can resolve it by only using `--testmon` to collect the tests which needs to be executed and then only running those, so we would do it in two steps: 1) filter test, 2) execute filtered tests. Now we do both of these in one go, and that seems not to work with `--testmon`.

max-models:
Moved to https://github.com/struphy-hub/struphy/pull/96

PR #65, opened 2025-10-27
Fixed redefinition of variables

### LINKED ISSUE #51: Redefinition of `PushVinViscousPotential` and other variables
Looks like `PushVinViscousPotential` is duplicated in the code

* https://github.com/struphy-hub/struphy/blob/devel/src/struphy/propagators/propagators_markers.py#L1781
* https://github.com/struphy-hub/struphy/blob/devel/src/struphy/propagators/propagators_markers.py#L1912

max-models:

Closes #51 

I deleted variables that were unused before redefinition according to `ruff check --select F811`.

https://docs.astral.sh/ruff/rules/redefined-while-unused/

The biggest change was the class `PushVinViscousPotential `, which seems to be written twice. I removed the first one, but please can someone check that the version that remains is correct?


max-models:
> ruff check --select F811

Resolved this in [08e9977](https://github.com/struphy-hub/struphy/pull/65/commits/08e9977db80c3064b64405a01d192ad09cc76977), the first one is 2D and second one is 3D

PR #66, opened 2025-10-28
37 use the GitHub container registry to store images

spossann:

Closes #...

**Core changes:**

None

**Model-specific changes:**

None

**Documentation changes:**

None


.gitlab/merge_request_templates/struphy_merge_request.md
max-models: I don't think we need to delete any of the `.gitlab` templates.

PR #69, opened 2025-10-29
Replaced `== None` with `is None`

### LINKED ISSUE #68: Replace `== None` with `is None`
See errors with:

```
ruff check --select E711
```

https://docs.astral.sh/ruff/rules/none-comparison/

At the same time, let's also fix these errors:

```
 1      E714    [*] not-is-test
 1      E721    [ ] type-comparison
 1      F722    [ ] forward-annotation-syntax-error
```

max-models:

Closes #68

According to [PEP 8](https://peps.python.org/pep-0008/#programming-recommendations), "Comparisons to singletons like None should always be done with is or is not, never the equality operators."

This PR fixes Ruff errors [E711](https://docs.astral.sh/ruff/rules/none-comparison/), [E714](https://docs.astral.sh/ruff/rules/not-is-test/), [E721](https://docs.astral.sh/ruff/rules/type-comparison/), and [F722](https://docs.astral.sh/ruff/rules/forward-annotation-syntax-error/).

### DIFF
--- a/src/struphy/feec/psydac_derham.py
+++ b/src/struphy/feec/psydac_derham.py
@@ -1304,7 +1304,7 @@ class Derham:
             neigh_inds = xp.array(neigh_inds)
 
             # only use indices where information is present to find the neighbours rank
-            inds = xp.where(neigh_inds is not None)
+            inds = xp.where(xp.not_equal(neigh_inds, None))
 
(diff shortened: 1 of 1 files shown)


max-models:
Replaced with https://github.com/struphy-hub/struphy/pull/90

PR #118, opened 2025-11-12
Auto install psydac on first struphy import

max-models:
- [x] Automatically install psydac on first import of struphy by moving all the psydac install logic from `compile.py` to `__init__.py`.
- [x] Remove redundant `struphy compile --status` commands

Try it out by removing the psydac dependency and then import struphy with

```
pip install .
pip uninstall psydac
python -c "import struphy"
```

This should install psydac since it isn't installed, and it isn't found in pyproject.toml


max-models:
This PR is not finished, but I think you can finish it @spossann. Just make sure python -c "import struphy" works both when `psydac` is and isn't present in the dependencies.

.github/actions/tests/models/action.yml
```diff
@@ -6,15 +6,13 @@ runs:
...
       run: |
-        struphy compile --status
         struphy test LinearMHD
         struphy test toy
         struphy test models 
         struphy test verification 
     - name: Model tests with  MPI
       shell: bash
       run: |
-        struphy compile --status
```
spossann: keep.

spossann (changes requested):

.github/actions/tests/models/action.yml
spossann: solved in [88947bf](https://github.com/struphy-hub/struphy/pull/118/commits/88947bf71360210864a2ee5dcf3fb744a55ac2d8)

.github/actions/tests/unit/action.yml
```diff
@@ -6,14 +6,12 @@ runs:
...
       shell: bash
       run: |
-        struphy compile --status
         struphy --refresh-models
         struphy test unit --mpi 2
 
     - name: Run unit tests
       shell: bash
       run: |
-        struphy compile --status
```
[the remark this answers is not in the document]
spossann: solved in [88947bf](https://github.com/struphy-hub/struphy/pull/118/commits/88947bf71360210864a2ee5dcf3fb744a55ac2d8)

spossann (approved):

PR #139, opened 2025-11-22
init .testmondata with non-MPI call

### LINKED ISSUE #138: .testmondata initialization
For tests with MPI, the `--testmon` flag leads to an INTERNALERROR and crashes when the `.testmondata` database does not yet exist.

Solution: run a simple non-MPI test before the first MPI test in a workflow to create the database.

spossann:

Closes #138

### DIFF
--- a/.github/workflows/test-PR-models-clones.yml
+++ b/.github/workflows/test-PR-models-clones.yml
@@ -74,6 +74,8 @@ jobs:
         shell: bash
         env:
           TESTMON_DATAFILE: ${{ github.workspace }}/.testmondata-clones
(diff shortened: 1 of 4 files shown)


max-models:
So this should resolve this error? https://github.com/struphy-hub/struphy/actions/runs/19595291176/job/56119131981?pr=129

max-models:
> > So this should resolve this error? https://github.com/struphy-hub/struphy/actions/runs/19595291176/job/56119131981?pr=129
> 
> Exactly.

Great, looks good to me!

PR #140, opened 2025-11-24
Repair struphy-install action

spossann:
* do not use venv in `struphy-install` action; prepare Python env instead
* try to make MacOS workflow actually do something
* clean up PR tests
* run PR tests on scheduled week-end pipeline


spossann:
The MacOs workflow now works: https://github.com/struphy-hub/struphy/actions/workflows/macos-latest.yml

PR #144, opened 2025-12-02
Enable TAE-EP simulations (fix several bugs)

bkna0327:
**Core changes:**
Allow `TAE-EP simulations` with devel.
 - Renew `CanonicalMaxwellian`.
 - Pass `n_cols_diag` and `n_cols_aux` to the class `Particles`.

**Model-specific changes:**

Bug fix in the model `LinearMHDDriftkineticCC`


src/struphy/kinetic_background/base.py
```diff
@@ -580,3 +580,233 @@ def add_perturbation(self) -> bool:
...
+            Shifted canonical toroidal momentum.
+
+        Returns
+        -------
+        A list[float] (background values) or a list[numpy.array] of the evaluated thermal velocities.
+        """
+        pass
+
+    @abstractmethod
+    def psic_to_eta(self, psci):
```
spossann:
```suggestion
    def psic_to_eta(self, psic):
```

src/struphy/kinetic_background/base.py
```diff
@@ -580,3 +580,233 @@ def add_perturbation(self) -> bool:
...
+        """
+
+        if isinstance(vth, xp.ndarray):
+            assert e.shape == vth.shape, f"{e.shape = } but {vth.shape = }"
+
+        out = 2.0 * xp.sqrt(e / xp.pi) / vth**3 * xp.exp(-e / vth**2)
+
+        return out
+
+    def __call__(self, *args):
```
spossann: Here it is not clear whether one evaluates at `eta` or at `psic`. I would opt for `eta` for consistency.

src/struphy/kinetic_background/maxwellians.py
```diff
@@ -245,14 +245,23 @@ def velocity_jacobian_det(self, eta1, eta2, eta3, *v):
...
+        assert isinstance(v[1], xp.ndarray)
+        assert eta1.shape == eta2.shape == eta3.shape == v[0].shape == v[1].shape
+        assert eta1.ndim == 1, "Input arguments must be a marker array."
+
+        etas = [
+            xp.concatenate(
+                (eta1[:, None], eta2[:, None], eta3[:, None]),
+                axis=1,
+            ),
+        ]
```
spossann: Is this equivalent to the deleted `vstack`?

src/struphy/kinetic_background/maxwellians.py
```diff
@@ -399,122 +396,41 @@ def check_maxw_params(self):
...
+        """Collection of factors multiplied onto the defined moments n and vth."""
         return self._moment_factors
 
     @moment_factors.setter
     def moment_factors(self, **kwargs):
         for kw, arg in kwargs:
             self._moment_factors[kw] = arg
 
-    def rc(self, psic):
+    def psic_to_eta(self, psic):
```
spossann: I propose the inverse: define a function `eta_to_psic` and evaluate everything at `eta`, for consistency with `KineticBackground`. This could be important when adding perturbations or MHD backgrounds.

spossann (changes requested):
Looks good! Have written some suggestions.

src/struphy/kinetic_background/maxwellians.py
bkna0327: Yes, it is, but wrapping with the list structure instead of a full copy of the array.

src/struphy/kinetic_background/base.py
```diff
@@ -580,3 +580,233 @@ def add_perturbation(self) -> bool:
...
+            vth = xp.moveaxis(vth, -1, 0)
+        else:
+            vth = vths
+
+        e = args[0]
+        res *= self.gaussian(e, vth=vth)
+
+        return res
+
+    def _evaluate_moment(self, psic, *, name: str = "n", add_perturbation: bool = None):
```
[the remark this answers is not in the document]
spossann: Please adapt the name to `rc` to be clear.

max-models:
> After implementing this, I realized that the performance degradation might be more significant than I initially expected. The current fix introduces multiple large 2D array computations (roughly number_of_particle_sizes × 6). Moreover, these calculations must be done every time the particle positions or velocities are updated when using the control variate method.

We should definitely finish the performance testing to make sure this doesn't happen 😅

bkna0327:
> After implementing this, I realized that the performance degradation might be more significant than I initially expected. The current fix  introduces multiple large 2D array computations (roughly number_of_particle_sizes × 6). Moreover, these calculations must be done every time the particle positions or velocities are updated when using the control variate method. 

This comment refers specifically to the most recent commit that I made in this MR, not to the overall approach in general ;)

spossann:
Changes have been merged in https://github.com/struphy-hub/struphy/pull/275

PR #147, opened 2025-12-10
Replace psydac -> feectools

### LINKED ISSUE #146: Install feectools
We want to get rid of the `.whl` install of `psydac-for-struphy` because it slows down our release process considerably. Instead, we plan to release it on PyPI under the name `feectools`, see https://github.com/struphy-hub/psydac-for-struphy/pull/60.

We thus need to replace `from psydac import ...` by `from feectools import ...` everywhere and also add `feectools` as a dependency.

spossann:

Closes #146 

- [x] set `feectools` as a dependeny once [this PR in psydac-for-struphy](https://github.com/struphy-hub/psydac-for-struphy/pull/60) has been merged.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -26,7 +26,7 @@ dependencies = [
     'numpy',
     'cunumpy',
     'pyccel>=2.0',
-    'feectools @ git+https://github.com/struphy-hub/psydac-for-struphy.git@feec-for-struphy',
+    'feectools',
     'scipy',
     'h5py',
     'matplotlib',


src/struphy/feec/psydac_derham.py
```diff
@@ -147,12 +147,19 @@ def __init__(
         self._polar_ck = polar_ck
 
         # derham sequence
+        try:
+            import psydac
+        except ModuleNotFoundError:
+            pass
+        use_feectools = True
```
spossann: It is enforced now.

max-models: Ok, but then why is `use_feectools` needed?

spossann: In the future it should be indented, so that we can also use psydac, but not now. This requires some more refactoring.

PR #152, opened 2025-12-11
13 move each model to its own file

max-models:
- [x] Move all models to its own file
- [x] Remove toy/fluid/kinetic/hybrid from the imports
- [x] Adapt the docs
- [x] Remove the refresh models command
- [x] Add `model_type` to each model, see [Added a model_type to each model](https://github.com/struphy-hub/struphy/pull/152/commits/21407864b25ca2e627055cbd56f4ae20ac15adff)
- [x] Add `profiling_tests` (merge https://github.com/struphy-hub/struphy/pull/127 first)
- [x] replace `struphy test ...` with the actual command, avoid sub-processes because they don't work with MPI

### DIFF
--- a/src/struphy/models/tests/default_params/test_models.py
+++ b/src/struphy/models/tests/default_params/test_models.py
@@ -9,57 +9,82 @@ from struphy.models.tests import utils_testing as ut
 @pytest.mark.models
 @pytest.mark.toy
 def test_toy(
+    test_profiling: bool = True,
     vrbose: bool = False,
     nclones: int = 1,
     show_plots: bool = False,
 ):
     for model in models_utils.get_models(model_type="Toy"):
-        ut.call_test(model=model(), verbose=vrbose)
+        ut.call_test(
+            model=model(),
+            test_profiling=test_profiling,
+            verbose=vrbose,
+        )
 
 
 @pytest.mark.models
 @pytest.mark.fluid
 def test_fluid(
+    test_profiling: bool = True,
     vrbose: bool = False,
     nclones: int = 1,
     show_plots: bool = False,
 ):
     for model in models_utils.get_models(model_type="Fluid"):
-        ut.call_test(model=model(), verbose=vrbose)
+        ut.call_test(
+            model=model(),
+            test_profiling=test_profiling,
+            verbose=vrbose,
+        )
 
 
 @pytest.mark.models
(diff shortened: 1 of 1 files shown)


spossann:
Hmm, does the error come from the struphy binary? It should not be called at all here, right?

https://github.com/struphy-hub/struphy/actions/runs/20329474755/job/58401285084#step:13:58

spossann:
> > > Hmm, does the error come from the struphy binary? It should not be called at all here, right?
> > > https://github.com/struphy-hub/struphy/actions/runs/20329474755/job/58401285084#step:13:58
> > 
> > 
> > I think it's called with `struphy test LinearMHD`, but for some reason, the subprocess pytest call fails before printing anything
> 
> Running the tests manually from within `src/struphy` works: https://github.com/struphy-hub/struphy/actions/runs/20330332144/job/58403886533?pr=152#step:13:5

Maybe the `libpath` to Struphy is wrong then, did you change it?

max-models:
> > > > > > Hmm, does the error come from the struphy binary? It should not be called at all here, right?
> > > > > > https://github.com/struphy-hub/struphy/actions/runs/20329474755/job/58401285084#step:13:58
> > > > > 
> > > > > 
> > > > > I think it's called with `struphy test LinearMHD`, but for some reason, the subprocess pytest call fails before printing anything
> > > > 
> > > > 
> > > > Running the tests manually from within `src/struphy` works: https://github.com/struphy-hub/struphy/actions/runs/20330332144/job/58403886533?pr=152#step:13:5
> > > 
> > > 
> > > Maybe the `libpath` to Struphy is wrong then, did you change it?
> > 
> > 
> > The libpath looks right, it's really confusing, actually.
> 
> One thing about the libpath: it seems struphy is installed in site-packages of the env, but what works is when you call the test from the repo source (thst would correspond to an editable install). Maybe you can try `install -e` and see if it works.
> 
> The libpath must always point to where struphy is actually installed.

The error in this PR is still a complete mystery to me. I am getting the same error on my PC (but not on my Mac). It seems to have something to do with the `mpirun` command because when I remove the `mpirun` like in [ce2c00e](https://github.com/struphy-hub/struphy/pull/152/commits/ce2c00e03207bd560568514141ae8b0bf55144a3) and then execute `struphy test Maxwell` it works fine.

PR #155, opened 2025-12-16
Install struphy on clusters

max-models:
- [x] Added private runners on TOK, Raven, Viper, and Pitagora.
- [x] Add script for loading the recommended module files for each cluster
- [x] Add CI job which installs the code on each cluster
- [x] Only run the CI job in a weekly scheduled pipeline (?)

The `.github/scripts/install_and_test.sh` script is executed only on the private runners, it installs, compiles and tests struphy. It can be executed manually too and should work on Raven/Viper/TOK/Pitagora

To simplify the installation on the clusters, I've added the following bash files in `setup/`:

- `modules.pitagora.sh`
- `modules.raven.sh`
- `modules.tok.sh`
- `modules.viper.sh`
- `modules.sh`

The `modules.MACHINE` contain the variables `MODULES_INTEL` and `MODULES_GCC` which lists the modules needed for intel/gcc on the current machine.

The `modules.sh` detects which machine we are on, and helps you load or show the modules needed.

```
# Usage:
#   ./modules.sh [load|display]
#     load    - Loads the required modules for the detected system/compiler
#     display - Prints the module load command for the detected system/compiler
```

So for example, if you want to load the GCC modules (which is default), you just type:

```
source ./setup/modules.sh load
# or
source ./setup/modules.sh load gcc
```

To load the intel modules, you type:

```
source ./setup/modules.sh load intel
```


max-models:
@spossann I think this PR can be reviewed now

spossann (approved):
This looks very nice! Did you test it on the machines?

max-models:
> This looks very nice! Did you test it on the machines?

Yes! This setup works and was tested on Raven, Viper and Pitagora!

max-models:
Also, in the future we could consider using something like this script for detecting the machine we are working on https://github.com/max-models/whereami

The idea of this small tool is to have one method across multiple projects to detect where code is running so the correct modules are loaded.

PR #159, opened 2026-01-11
Generate parameter files with AST

max-models:
Trying out some code generation, I think it could be useful at some point. Can't exactly justify it for a specific purpose. Maybe it could help with parameter file generation with some LLM or exporting a runnable script from checking boxes in a GUI...

For example, this produces the same parameters as `struphy params Maxwell`:

```
python generate_maxwell_ast.py > params_Maxwell.py
python params_Maxwell.py 
```


max-models:
Once #186 is merged, I think this MR will become easier to finish.

PR #162, opened 2026-01-16
161 add binning of higher order moments of f and delta f

P-Attapon:
Add options to bin **current density** and **energy tensor** in `BinningPlot` instead of only particle density

**Core changes:**

1. Implement weights calculation based on type of quantity to calculate.
2. For saving data, the type of quantity is attached to the slice name.  
Example: slice = "e1_v1", output_quantity = "current_density_v1"
=> `simdata.f["kinetic_ions"]["e1_v1_current_density_v1"]`
slice = "e1_v1", output_quantity = "particle"
=>` simdata.f["kinetic_ions"]["e1_v1"]`
3.  We make the signature and docstring of `Particles` base class appear in VScode, when an instance of Particle class is created:
    * new abstract method `__post_init__(self)` is called at the end of base `__init__(self)`
    * The handling of the default column numbers had to be changed
    * Move docstring from class to `__init__`

**Model-specific changes:**

None
Older parameter files can be compiled directly without any changes

**Documentation changes:**

1. Create a new set of Literal `BinningOutput`. This Literal is used to determine the quantity to bin and the axis of velocity for weights in binning calculation. The three quantities are `current_density`, `energy_tensor`, and `particle`.

### DIFF
--- a/src/struphy/pic/tests/test_binning.py
+++ b/src/struphy/pic/tests/test_binning.py
@@ -1028,10 +1028,10 @@ def test_binning_current_6D_full_f(mapping, show_plot=False):
         LoadingParameters,
         WeightsParameters,
     )
+
     # ==============================================================
     # ===== Setup default parameters =====
     # ==============================================================
-
     # test weights
     amp_n = 0.1
     mode_number = 2
@@ -1136,10 +1136,10 @@ def test_binning_current_6D_full_f(mapping, show_plot=False):
     # ==== Test different direction of perturbation and current ====
     # ==============================================================
 
-    pert = perturbations.ModesCos(ls = (mode_number,), comp = 0)
-    maxwellian = Maxwellian3D(u2 = (u_norm, pert))
+    pert = perturbations.ModesCos(ls=(mode_number,), comp=0)
+    maxwellian = Maxwellian3D(u2=(u_norm, pert))
 
-    loading_params = LoadingParameters(Np=Np, moments=[0.0,u_norm,0.0,1.0,1.0,1.0])
+    loading_params = LoadingParameters(Np=Np, moments=[0.0, u_norm, 0.0, 1.0, 1.0, 1.0])
     boundary_params = BoundaryParameters()
 
     particles = Particles6D(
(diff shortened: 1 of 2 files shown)


src/struphy/models/base.py
```diff
@@ -1115,8 +1118,14 @@ def initialize_data_output(self, data: DataContainer, size):
 
                 # binning plot data
                 for bin_plot in species.binning_plots:
-                    key_f = os.path.join(key_spec, "f", bin_plot.slice)
-                    key_df = os.path.join(key_spec, "df", bin_plot.slice)
+
+                    # define slice name with binning quantity
+                    slice, output_quantity = bin_plot.slice, bin_plot.output_quantity
+                    if output_quantity != "particle":
```
spossann: You can remove this `if` statement, treat all the same.

src/struphy/pic/base.py
```diff
@@ -1936,12 +1936,22 @@ def binning(
         _n = len(components)
         slicing = components + [False] * (self.markers.shape[1] - _n)
 
+        # determine histogram weights multiplier
+        multiplier = 1
+        # determine type of output quantity
+        if output_quantity != "particle":
```
spossann: Remove this `if`.

src/struphy/pic/base.py
```diff
@@ -1936,12 +1936,22 @@ def binning(
...
+            quantity, v_axis = output_quantity.rsplit(sep = "_", maxsplit = 1)
+            v_axis = [int(char) - 1 for char in v_axis if char.isnumeric()] # convert dimension axis to index
```
spossann:
```suggestion
            v_axis = [int(char) - 1 for char in v_axis] # convert dimension axis to index 
```

spossann (changes requested):
This look pretty good! I have added some minor suggestions and the folllowing points:

1. Please add unit test for the new binning outputs in `pic/tests/test_binning`. You can write new tests such as `def test_binning_current_6D_full_f(mapping, show_plot=False):`  and adapt the following code accordingly:
```
# =========================================
    # ===== Test cosine in eta1 direction =====
    # =========================================
    # test weights
    amp_n = 0.1
    l_n = 2
    pert = perturbations.ModesCos(ls=(l_n,), amps=(amp_n,))
    maxwellian = Maxwellian3D(n=(1.0, pert))

    particles = Particles6D(
        loading_params=loading_params,
        boundary_params=boundary_params,
        domain=domain,
        background=maxwellian,
    )
    particles.draw_markers()
    particles.initialize_weights()

    e1_bins = xp.linspace(0.0, 1.0, 200, endpoint=True)
    de = e1_bins[1] - e1_bins[0]

    binned_res, r2 = particles.binning(
        [True, False, False, False, False, False],
        [e1_bins],
    )

    e1_plot = e1_bins[:-1] + de / 2

    ana_res = 1.0 + amp_n * xp.cos(2 * xp.pi * l_n * e1_plot)

    if show_plot:
        plt.plot(e1_plot, ana_res, label="Analytical result")
        plt.plot(e1_plot, binned_res, "r*", label="From binning")
        plt.title(r"Full-$f$: Cosine in $\eta_1$-direction")
        plt.xlabel(r"$\eta_1$")
        plt.ylabel(r"$f(\eta_1)$")
        plt.legend()
        plt.show()

    l2_error = xp.sqrt(xp.sum((ana_res - binned_res) ** 2)) / xp.sqrt(xp.sum((ana_res) ** 2))

    assert l2_error <= 0.02, f"Error between binned data and analytical result was {l2_error}"
```
2. Please fix the unit test https://github.com/struphy-hub/struphy/actions/runs/21075207400/job/60615217672?pr=162#step:10:6877

3. You have to run `struphy format all` to format the code correctly, such that the CI tests pass.

spossann (changes requested):
In order to be able to format you need to install as a developer `pip install -e .[dev]`.

PR #168, opened 2026-02-04
Add a public API

max-models:
This is my take on defining a public API using `__init__.py` files, it allows for importing all the neccesary components as follows:

```python
from struphy import EnvironmentOptions, BaseUnits, Time
from struphy import domains
from struphy import equils
from struphy import grids
from struphy import DerhamOptions
from struphy import FieldsBackground
from struphy import perturbations
from struphy import maxwellians
from struphy import (
    LoadingParameters,
    WeightsParameters,
    BoundaryParameters,
    BinningPlot,
    KernelDensityPlot,
)
from struphy import main

from struphy.models import Vlasov
```

Check the file `params_Vlasov.py` for a working example.

# Todo

- [x] Put modules for the public API ion `api/`, this should not contain any code, just `__init__.py` files
- [x] Import the modules from the highest level and expose everything to the user
- [ ] ~~(maybe) rename `models` to `plasma_models` and then later in `struphy.api.models` we import each specific model (removes clash with existing `struphy.models` module)~~
- [ ] ~~(maybe) hide the rest of the code from the user by moving it to `_internal` or similar, we should check what other projects do here.~~

### DIFF
--- a/src/struphy/__init__.py
+++ b/src/struphy/__init__.py
@@ -12,8 +12,6 @@ from struphy.api.particles import (
 )
 from struphy.api.perturbations import perturbations
 
-# from struphy.api import models # This is redundant
-
 __all__ = [
     "domains",
     "equils",
@@ -30,5 +28,4 @@ __all__ = [
     "KernelDensityPlot",
     "DerhamOptions",
     "FieldsBackground",
-    # "models", # Redundant since struphy.models already points to this module
 ]


src/struphy/api/options/__init__.py
spossann: `Units` should not be part of the api, but this we will change once #167 is merged.

src/struphy/api/pic_utilities/__init__.py
spossann: Instead of `api/pic_utilities` I suggest to call this `api/particles`, since it is not just for pic, but all particle methods's api.

spossann (approved):
This is really nice! In what follows maybe we could also make a folder `core` or `_core` for base classes etc.

params_Vlasov.py
max-models: This should be deleted before merging.

src/struphy/api/io_options/__init__.py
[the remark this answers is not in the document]
max-models: Fix in [25fbda4](https://github.com/struphy-hub/struphy/pull/168/commits/25fbda40debd96f6cbccfcd9a83e5ed01dd76bdc)

src/struphy/api/options/__init__.py
max-models: Fixed in [8c42bf5](https://github.com/struphy-hub/struphy/pull/168/commits/8c42bf55f5b1041e5f1c2c65a4b8b56a42330b35)

PR #177, opened 2026-02-13
Simulation class

### LINKED ISSUE #169: Run a model with `model.run()`
The way the model is executed in the python parameter files is a bit counter-intuitive to me.

As a user of the code, I would assume the the model itself would have a `run` method, not that we have to import `main` and pass the model as a parameter.

Current method:

```python
if __name__ == "__main__":
    main.run(
        model,
        params_path=__file__,
        env=env,
        base_units=base_units,
        time_opts=time_opts,
        domain=domain,
        equil=equil,
        grid=grid,
        derham_opts=derham_opts,
        verbose=verbose,
    )
```

Suggested change:

```python
if __name__ == "__main__":
    model.run(
        params_path=__file__,
        env=env,
        base_units=base_units,
        time_opts=time_opts,
        domain=domain,
        equil=equil,
        grid=grid,
        derham_opts=derham_opts,
        verbose=verbose,
    )
```

This would just require moving the content in `main.py` to the models baseclass.

spossann:

Closes #169

**Core changes:**

1. New class `Simulation` inherits the generic `SimulationBase`, both are in the new folder `struphy/simulations/`. The most import methods are:
* `Simulation.run()`
* `Simulation.pproc()` 
* `Simulation.load_plotting_data()`
* `Simulation.spawn_sister()` (my new favorite!)

These are tested in the new tutorials: https://github.com/struphy-hub/struphy-tutorials/tree/use-species-properties

The file `main.py` has been deleted.

The `Simulation` takes a model as input. Other API classes are passed as well (see tutorials). 
The model is viewed as everything related to the PDE, i.e. its variables, initial conditions etc. The simulation deals with the rest (geometry, derham, environment etc.)

Some important changes to the logic: The model does not have access to `derham`, `mass_ops` etc. anymore, these can be called from `Propagator` when needed. Solves that need to happen before the time stepping (like initial Poisson solves) are moved to `model.allocate_helpers()`.

2. The default launch file has been improved. See Tutorial 2 or test with `struphy params MODEL`. Thus, the available files in the submodule  struphy-parameters must be adapted (TODO).

3. Several new classes have been introduced for post processing and plotting data, see `post_processing_tools.py`. The most important ones are `PostProcessor` and `PlottingData`. Dictionaries in the plotting data have been replaced by classes. Many classes now feature the `__repr__` dunder for customized printing.

4. `verbose: bool = False` has been added everywhere in the simulation methods. The handling of verbosity will need some more improvements in the future though.

5. Disable the `ruff format --check` [e0544eb](https://github.com/struphy-hub/struphy/pull/177/commits/e0544ebcd158263bc1ef29069102db6de94fbc2c)

6. Introduce class properties in Particles classes [1449b5e](https://github.com/struphy-hub/struphy/pull/177/commits/1449b5e352bb6fd427db2e95092534a71c6ddee3)

**API changes:**

New classes exposed: `Simulation`, `PostProcessor` and `PlottingData`.

**Model-specific changes:**

None

**Documentation changes:**

The Quickstart has been adapted to the new classes.

**Things left TODO:**

- [x] Adapt submodule struphy-parameters
- [x] Add docstrings

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -34,6 +34,7 @@ dependencies = [
     'vtk',
     'tqdm',
     'argcomplete',
+    'plotly',
     'pytest',
     'pytest-mpi',
     'pytest-testmon',
@@ -65,7 +66,6 @@ dev = [
     "tomli-w",
     "kaleido",
     "pandas",
-    "plotly",
 ]
 mpi = [
     'mpi4py',


spossann:
> What is the point of passing `params_path=__file__,`? Is it just to copy the parameter file to the output directory?

Yes, I think this should be done in any case, even if the file uses other custom files.

> 
> If so, I think we can deduce the file which calls Simulation with
> 
> ```
> frame = inspect.stack()[1]
> caller_file = frame.filename
> ```

Ok, if we can get rid of the argument `params_path` this is good. We need to be able to also treat the case `params_path = None` (for notebook runs for example).

> 
> But in any case, I feel like this approach is a bit misguided, because there is nothing special about the _file_ which generates the simulation. I could very much see myself using a workflow with multiple files containing my typical setups and importing struphy objects from those and combining them into a `Simulation` in a short file.
> 
> For example:
> 
> ```python
> from struphy.models import Maxwell
> from my_setup import my_special_domain
> 
> model = Maxwell()
> sim = Simulation(
>     model=model,
>     domain= my_special_domain,
> )
> ```
> 
> Passing `params_path` here makes no sense, since the "params" are in multiple files.
> 
> Instead, what I think we should do is to _generate_ a new parameter file from the `Simulation` class, since everything needed for a simulation should already be contained in it. So I would suggest we add a `generate_params()` method to `Simulation`. This would of course be a bit more complex than generating default parameters because we have to check which arguments were defaults and which were not.

How do we get the Simulation object from a previous simulation? If we can, that would be nice.

PR #183, opened 2026-02-25
182 measure gauss law violation

### LINKED ISSUE #182: Measure Gauss law violation
Implement new feature to measure violation of Gauss's law

$$\nabla \cdot \mathbf{E} = \frac{\rho}{\epsilon_0} \Rightarrow \text{error(t)} = \max(|\mathbf{G}^T\mathbf{M}^1\mathbf{e} -  \mathbf{f}^p|)$$

P-Attapon:

Implement new feature to measure violation of Gauss's law

$$\nabla \cdot \mathbf{E} = \frac{\rho}{\epsilon_0} \Rightarrow \text{error(t)} = \max(|\mathbf{G}^T\mathbf{M}^1\mathbf{e} -  \mathbf{f}^p|)$$

**Core changes:**
Add measurement of gauss error to VlasovMaxwellOneSpecies:  
`src/struphy/models/vlasov_maxwell_one_species.py`

**Documentation changes:**
1. New attribute `measure_gauss_error` to VlasovMaxwellOneSpecies model
2. new line in default parameter file of VlasovMaxwellOneSpecies


src/struphy/models/vlasov_maxwell_one_species.py
```diff
@@ -156,6 +156,26 @@ def __init__(self):
...
+        # property to measure violation of gauss law from control variate
+        self.measure_gauss = False
+
+    def measure_gauss_error(self, measure: bool = False):
+        """
+        Record gauss law violation during simulaiton:
+
+        .. math::
+            error(t) = \max(
+                |\mathbf{G}^T\mathbf{M}^1\mathbf{e} -  \mathbf{f}^p|
```
spossann:
```suggestion
                |\mathbb{G}^T\mathbb{M}^1\mathbf{e} -  \mathbf{f}|
       
    where :math:`\mathbf f = (f_i)_i` with the accumulation vector :math:`f_i = \frac 1N \sum_p w_p \Lambda_i^0(\eta_p)`.                       
```

src/struphy/models/vlasov_maxwell_one_species.py
```diff
@@ -156,6 +156,26 @@ def __init__(self):
...
+                )
+
+        If control_variate = False, then the error should always be zero
```
spossann: This is not true actually, it depends on the discretization. Our algo does not conserve the Gauss law exactly, this would require a symplectic algo, but we have an energy conserving one.

src/struphy/models/vlasov_maxwell_one_species.py
```diff
@@ -213,6 +233,28 @@ def allocate_helpers(self, verbose: bool = False):
         # reset particle weights
         particles.weights = particles.weights_at_t0.copy()
 
+    def calculate_gauss_error(self, e):
```
spossann: The coefficients "e" do not need to be passed here, they can be retrieved from the model:
```
e = self.em_fields.e_field.spline.coeffs
```

src/struphy/models/vlasov_maxwell_one_species.py
```diff
@@ -213,6 +233,28 @@ def allocate_helpers(self, verbose: bool = False):
...
+        # control variate method
+        particles = self.kinetic_ions.var.particles
+
+        charge_accum0 = AccumulatorVector(
+            particles,
+            "H1",
+            Pyccelkernel(accum_kernels.charge_density_0form),
+            Propagator.mass_ops,
+            Propagator.domain.args_domain,
+        )
```
spossann: This should be instantiated only once in the init of the model, not at every call.

src/struphy/models/vlasov_maxwell_one_species.py
```diff
@@ -213,6 +233,28 @@ def allocate_helpers(self, verbose: bool = False):
...
+        charge_accum0 = xp.array(charge_accum0.vectors[0].toarray())
```
spossann: Here you need to accumulate by calling the object: `charge_accum0()`. After that, thes result will be in `charge_accum0.vectors[0]`. You do not need to convert toarray at this point.

src/struphy/models/vlasov_maxwell_one_species.py
```diff
@@ -213,6 +233,28 @@ def allocate_helpers(self, verbose: bool = False):
...
+
+        # non control variate method
+        op = Propagator.derham.grad.T @ Propagator.mass_ops.M1
+        charge_accum1 = op.dot(e)
+        charge_accum1 = charge_accum1.toarray()
```
spossann: In an MPI run with multiple ranks, the `toarray` method will give you an array of global size, but with nonzero elements only at the places that are available on the current rank. We should compute the local differences and then allreduce them to rank zero, which then takes the max value of all ranks:
```
rhs = charge_accum0.vectors[0]
lhs = op.dot(e)
loc_residual = xp.max(xp.abs(rhs.toarray() - lhs.toarray()))
```
You can the communicate the local result by using the method `Particles._gather_scalar_in_subcomm_array` and subsequently `Particles._gather_scalar_in_intercomm_array`. Maybe you have to modify these in order to accept floats.

src/struphy/models/vlasov_maxwell_one_species.py
```diff
@@ -213,6 +233,28 @@ def allocate_helpers(self, verbose: bool = False):
...
+
+        # take maximum from difference of two methods
+        residual = xp.max(xp.abs(charge_accum0 - charge_accum1))
+        return residual
```
spossann: return the max residual after MPI communication.

src/struphy/models/vlasov_maxwell_one_species.py
```diff
@@ -213,6 +236,29 @@ def allocate_helpers(self, verbose: bool = False):
...
+
+        # calculate local residual of local MPI rank
+        loc_residual = xp.max(xp.abs(lhs.toarray() - rhs.toarray()))
+
+        # return the maximum residual across all MPI rank
+        subcom_residual = xp.empty(shape=particles.mpi_size, dtype = float)
+        intercom_residual = xp.empty(shape=particles.num_clones, dtype = float)
+
+        particles._gather_scalar_in_subcomm_array(scalar=loc_residual, out=subcom_residual)
+        particles._gather_scalar_in_intercomm_array(scalar=loc_residual, out=intercom_residual)
```
P-Attapon: I am not sure about this line because particles.clone_config seems to be False and the values across ranks are not allocated into one array

src/struphy/models/vlasov_maxwell_one_species.py
```diff
@@ -189,6 +210,8 @@ def allocate_helpers(self, verbose: bool = False):
             Propagator.mass_ops,
             Propagator.domain.args_domain,
         )
+        # store initial charge accumulation
+        self._charge_accum0 = charge_accum
```
P-Attapon: charge_accum0 is not calculated during __init__ but stored from this value

spossann: fixed in [b40903d](https://github.com/struphy-hub/struphy/pull/183/commits/b40903d75e42759888957d1e82f0f38d758316af)[b40903d](https://github.com/struphy-hub/struphy/pull/183/commits/b40903d75e42759888957d1e82f0f38d758316af)

src/struphy/models/vlasov_maxwell_one_species.py
spossann: fixed in [b40903d](https://github.com/struphy-hub/struphy/pull/183/commits/b40903d75e42759888957d1e82f0f38d758316af)[b40903d](https://github.com/struphy-hub/struphy/pull/183/commits/b40903d75e42759888957d1e82f0f38d758316af)

src/struphy/models/vlasov_maxwell_one_species.py
spossann: fixed in [c925b6e](https://github.com/struphy-hub/struphy/pull/183/commits/c925b6e098460e83415dbd2870d7cf5406c347bf)[c925b6e](https://github.com/struphy-hub/struphy/pull/183/commits/c925b6e098460e83415dbd2870d7cf5406c347bf) and [d601f05](https://github.com/struphy-hub/struphy/pull/183/commits/d601f0587c824f4f8f6d9f6fa223f9c4dcfadc64)[d601f05](https://github.com/struphy-hub/struphy/pull/183/commits/d601f0587c824f4f8f6d9f6fa223f9c4dcfadc64)

spossann (approved):

PR #185, opened 2026-02-28
Version 3.0.4

spossann:


spossann:
@max-models We see the Simulation class commit from the rebase, should we keep that? Instead, I guess we could do a merge and have a merge commit, what do you think for future releases?

max-models:
> @max-models We see the Simulation class commit from the rebase, should we keep that? Instead, I guess we could do a merge and have a merge commit, what do you think for future releases?

I think it's ok for now.

What I think we should do in the future is that we should not change the dependencies between devel and main. We should always use upper bounds on all the dependencies, but we should have write a simple tool that checks if any of the dependencies have a newer version. Maybe we should label the dependencies as "fixed" or "auto" with a comment and then the tool should only update the ones which are "auto".

Then we just update the dependencies from time to time to make sure we are up to date with the latest versions. There might even be some Github action for this.

max-models (approved):

PR #186, opened 2026-02-28
Add `to_dict` and `from_dict`methods to the Simulation class

max-models:

- Add `to_dict` and `from_dict`methods to the Simulation class and other parameter and core classes to enable serialization and deserialization of simulation configurations and objects. This allows saving, loading, and transferring simulation state and parameters in a structured way.
- Add `__eq__` method to the Simulation class

Example:

```python
from struphy import Simulation
from struphy.models import Maxwell

sim1 = Simulation(model=Maxwell())
sim2 = Simulation.from_dict(sim1.to_dict())

assert sim1 == sim2
```


src/struphy/fields_background/base.py
```diff
@@ -306,6 +306,30 @@ def u_cart_2(self, *etas, squeeze_out=False):
     def u_cart_3(self, *etas, squeeze_out=False):
         return self.u_cart(*etas, squeeze_out=squeeze_out)[0][2]
 
+    @abstractmethod
+    def to_dict(self) -> dict:
+        pass
```
spossann: Why not make this a common method:
```
    def to_dict(self) -> dict:
        return {
            "type": self.__class__.__name__,
            "params": self.params,
        }
```

src/struphy/fields_background/equils.py
```diff
@@ -163,6 +163,16 @@ def gradB_xyz(self, x, y, z):
...
+    def to_dict(self) -> dict:
+        return {
+            "type": "HomogenSlab",
+            "params": self.params,
+        }
+
+    @classmethod
+    def from_dict(cls, dct: dict) -> "HomogenSlab":
+        return cls(**dct["params"])
+
```
spossann: We could use the common methods from the base class.

src/struphy/fields_background/tests/test_numerical_mhd_equil.py
spossann, line 1: Let's add a small test to `test_mhd_equils.py`:
```
    # to-from-dict
    eq_dict = eq_mhd.to_dict()
    print("\neq_dict:")
    for k, v in eq_dict.items():
        print(f"{k} = {v}")
    eq_mhd_from_dict = eq_mhd.__class__.from_dict(eq_dict)
    print(f"{eq_mhd_from_dict = }")
    # assert eq_mhd_from_dict == eq_mhd
```
We could un-comment the last line if we had an `__eq__` also for equils.

src/struphy/geometry/base.py
spossann, line 1: Maybe add a similar unit test as I suggested for equils, using an `__eq__` for domains.

spossann (commented):
This will be very useful in the future! I think we can, in a next PR, replace the saving of the parameter file: https://github.com/struphy-hub/struphy/blob/4aa8f5097d3f02bf370f6d1b4cceeedb9566aa39/src/struphy/simulation/sim.py#L206

This will require also to replace `import_parameters_py` by loading the dict: https://github.com/struphy-hub/struphy/blob/4aa8f5097d3f02bf370f6d1b4cceeedb9566aa39/src/struphy/io/setup.py#L16

src/struphy/fields_background/base.py
max-models: Yes that's a better idea! I wasn't sure at first if all of them saved the params the same way, but since they do we can move everything to the baseclass!

max-models: Fixed in [a6a1bd1](https://github.com/struphy-hub/struphy/pull/186/commits/a6a1bd115b58e074ea111486981682ff478351c2)

src/struphy/fields_background/equils.py
max-models: Fixed in [a6a1bd1](https://github.com/struphy-hub/struphy/pull/186/commits/a6a1bd115b58e074ea111486981682ff478351c2)

src/struphy/fields_background/tests/test_numerical_mhd_equil.py
max-models: Fixed in [c9fdea1](https://github.com/struphy-hub/struphy/pull/186/commits/c9fdea12809519aed83dbe61ef9ecef70e262fee)

src/struphy/geometry/base.py
max-models: Fixed in [87196b6](https://github.com/struphy-hub/struphy/pull/186/commits/87196b62888bcd384e51713813b8a95da75c64a8)

spossann (approved):

PR #187, opened 2026-02-28
Html docstrings

spossann:
* New classmethod `StruphyModel.info()` that shows the html docstring [(see for instance updated tutorials 01 and 05)](https://github.com/struphy-hub/struphy-tutorials/tree/add-model-info)
* Helper methods for translating rst to markdown and html
* The conventional rst docstring now appears under `__doc_rst__`, which is processed by Sphinx. A custom
extension has been written for this.

In practice, we only write the `__doc_rst__` string and then auto-translate to html, which becomes the class docstring under the class name (this is seen by Pylance).

Models with improved docstrings so far:

- [x] Maxwell
- [x] VlasovAmpereOneSpecies
- [x] LinearMHD


src/struphy/utils/docstring_converter.py
```diff
@@ -0,0 +1,192 @@
...
+
+    If a class has a __doc_rst__ attribute, this converts it to HTML
+    and sets it as the class docstring for VS Code display.
+    """
+    if hasattr(cls, "__doc_rst__") and cls.__doc_rst__:
+        # Convert RST to HTML
+        html_doc = rst_to_html(cls.__doc_rst__)
+        # Set as the main docstring
+        cls.__doc__ = html_doc
+    return cls
```
max-models, line 845:
```suggestion
    return cls

if __name__ == "__main__":
    import argparse

    from struphy.models.utils import get_model_by_name
    from struphy.utils.docstring_converter import render_docstring

    parser = argparse.ArgumentParser()
    parser.add_argument(
        "model_name",
        type=str,
        help="Name of the model to convert docstring for",
    )
    parser.add_argument(
        "--use-rst",
        action="store_true",
        help="If set, use the RST version of the docstring",
    )

    args = parser.parse_args()

    model = get_model_by_name(args.model_name)

    docstring = render_docstring(model, use_rst=args.use_rst)

    print(docstring.data)
```

max-models: I wasn't really sure of how to use the functions in `src/struphy/utils/docstring_converter.py`, maybe we can add something like this to make it easier to use the script?

src/struphy/utils/docstring_converter.py
```diff
@@ -0,0 +1,192 @@
...
+            # It's HTML
+            return HTML(doc_text)
+        else:
+            # Plain text or RST, show as is
+            return Markdown(doc_text)
+    else:
+        return Markdown("*No docstring available*")
+
+
+def auto_convert_docstring(cls):
```
max-models, line 833: How do you use this? It doesn't actually change the text of the docstring in the model file I guess, just the `__doc__` attribute.

max-models:
By the way, this is a really nice upgrade! It makes the docstrings much more readable in pylance, although some of the characters are really small compared to compiled latex...

src/struphy/utils/docstring_converter.py
spossann: The function `render_docstring` was actually obsolete, its contents went into `StruphyModel.info()`. The `__doc__` is now auto generated upon module import via a decorator function. Pylance does not see the whole docstring, but only a simple intro sentence. The `__doc_rst__` has been moved to the file end.

src/struphy/utils/docstring_converter.py
spossann: It is now a decorator function, see [e936151](https://github.com/struphy-hub/struphy/pull/187/commits/e936151fd6dd06fd53826ab053e96896d04737a8)

max-models (dismissed):

src/struphy/simulation/sim.py
```diff
@@ -204,7 +204,7 @@ def __init__(
         )
 
         # save parameter file
-        if self.rank == 0 and verbose:
```
max-models, line 207: Hehe, interesting to see that this didn't cause any errors at all in my PR!

max-models (approved):

PR #193, opened 2026-03-04
Generate parameter files from Simulation class

### LINKED ISSUE #189: Generate parameter files from dict
Currently, we can import and export all simulation parameters with the `to_dict()` and `from_dict()` methods.

What we still can not do is to generate a python script that imports all needed components, builds the simulation object and runs it.

I think we should proceed with the following steps:

1. Add `__repr__` methods for all the objects to which we added to/from dict

2. Find a way to remove default parameters from the string returned by `__repr__`

So for example if we have this class

```
class Point:
    def __init__(self, x = 0, y = 0):
        self.x = x
        self.y = y
```
then the standard behavior of `__repr__` would be to return `Point(x = 0, y = 0)`, but maybe we need a method to return just `Point()` if x=0 and y=0. Otherwise our generated python scripts will be maximally populated with all the default arguments for every class.

3. Find a way to only import what is actually needed in the code. One very simple option would actually be to import everything, and then remove the unused imports with `ruff`... Otherwise we just have to include some logic in the script generation.

To make the imports cleaner, I think we should resolve https://github.com/struphy-hub/struphy/issues/170, I think we can then just rename the remaining modules, update the API a bit and then the imports will be cleaner. For example, we currently can not do `from struphy.domains import Cuboid`, instead we have to do `from struphy import domains; dom = domains.Cuboid()` or `from struphy.geometry.domains import Cuboid` which is a bit inconsistent with the imports of the other imports. Currently I think there are a few modules that should be exposed to the user: domains, equils, grids, perturbations and models.

max-models:
This PR is based on the idea that `__repr__` should [return a printable representation of an object](https://www.geeksforgeeks.org/python/python-__repr__-magic-method/). The goal is for all objects in the API to have this method, so that a simulation setup can be reproduced using a new `generate_script()` or `save_script(filepath)` method.

- Added `__repr__` and `__repr_no_defaults__` methods to most classes in the API. This allows us to create a repr of the instance without cluttering it with default parameters.
- Renamed a few old `__repr__` methods that would fit better as `__str__`
- Updated tutorials by using `print(x)` instead of `x`
- Added `generate_script()` and `save_script(filepath)` methods to the Simulation class
- The generated strings are formatted with ruff, which also removes unused imports. Therefore, ruff is moved from a dev dependency to a normal dependency. This is the easiest way to use a consistent formatting for the generated scripts.
- Added test to compare the script generation from two simulations.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -39,6 +39,7 @@ dependencies = [
     'pytest',
     'pytest-mpi',
     'pytest-testmon',
+    'ruff>=0.15.0',
     'line_profiler',
     'scope-profiler>=0.1.7',
 ]
@@ -60,7 +61,6 @@ dev = [
     "pylint",
     "ssort",
     "add-trailing-comma",
-    "ruff>=0.15.0",
     "pre-commit",
     "nbstripout",
     "tabulate",
--- a/src/struphy/fields_background/base.py
+++ b/src/struphy/fields_background/base.py
@@ -7,7 +7,11 @@ from matplotlib import pyplot as plt
 from pyevtk.hl import gridToVTK
 
 from struphy.geometry.base import Domain
-from struphy.utils.utils import __class_with_params_repr_no_defaults__, __dataclass_repr_no_defaults__
+from struphy.utils.utils import (
+    __class_with_params_repr_no_defaults__,
+    __dataclass_repr_no_defaults__,
+    all_class_params_are_default,
+)
 
 
 class FluidEquilibrium(metaclass=ABCMeta):
@@ -113,6 +117,10 @@ class FluidEquilibrium(metaclass=ABCMeta):
 
     def __repr_no_defaults__(self):
         return __class_with_params_repr_no_defaults__(self)
+
+    @property
+    def is_default(self):
+        return all_class_params_are_default(self)
 
     ###########################
     # Vector-valued callables #
--- a/src/struphy/fields_background/base.py
+++ b/src/struphy/fields_background/base.py
@@ -7,6 +7,7 @@ from matplotlib import pyplot as plt
 from pyevtk.hl import gridToVTK
 
 from struphy.geometry.base import Domain
+from struphy.utils.utils import __class_with_params_repr_no_defaults__, __dataclass_repr_no_defaults__
 
 
 class FluidEquilibrium(metaclass=ABCMeta):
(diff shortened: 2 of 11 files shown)


max-models:
Example:

```python
from struphy import Simulation

from struphy.models import Maxwell
model = Maxwell()

# Simulation object
sim = Simulation(
    model=model,
    params_path=__file__,
)

print(sim.generate_script())
```

generates the following output:

```python
from struphy import (
    BaseUnits,
    DerhamOptions,
    EnvironmentOptions,
    FieldsBackground,
    Simulation,
    Time,
    domains,
    equils,
    grids,
    perturbations,
)

# ---------------------
# Instance of the model
# ---------------------

from struphy.models import Maxwell

model = Maxwell()
env = EnvironmentOptions()
base_units = BaseUnits()
time_opts = Time()
domain = domains.Cuboid()
grid = grids.TensorProductGrid()
derham_opts = DerhamOptions()


sim = Simulation(
    model=model,
    params_path="/Users/max/git_repos/larmor-dev/struphy/params_Maxwell.py",
    env=env,
    base_units=base_units,
    time_opts=time_opts,
    domain=domain,
    grid=grid,
    derham_opts=derham_opts,
)
sim.run()
```

The last step will be to also remove the default parameters from the arguments in `Simulation` and then remove unused imports.

max-models:
Update:

**In**:

```python
from struphy import Simulation, Time

from struphy.models import Maxwell
model = Maxwell()
time_opts = Time(dt=0.1)

sim = Simulation(
    model=model,
    time_opts=time_opts,
    params_path=__file__,
)

print(sim.generate_script())
```

**Out**:

```python
from struphy import (
    Simulation,
    Time,
)

from struphy.models import Maxwell

model = Maxwell()
time_opts = Time(dt=0.1)

sim = Simulation(
    model=model,
    time_opts=time_opts,
    params_path="/Users/max/git_repos/larmor-dev/struphy/params_Maxwell.py",
)
```

spossann (approved):
This looks very good! We should probably use the new __repr__'s also to generate default parameter files.

max-models:
> This looks very good! We should probably use the new **repr**'s also to generate default parameter files.

Definitely, but we can do this in a separate PR!

The default parameter files are still a bit more complex, because in some cases the state of the `model` class is updated after initialization. I think we could try to optionally move all of that logic to the init.

PR #195, opened 2026-03-05
Add `show_3d` and `create_geometry_mesh` methods to domain class

max-models:
- Added a few new methods for viewing and exporting 3D views of the domains
  - `create_geometry_mesh`: Generates a pyvista mesh
  - `show_3d`: Displays the 3D geometry using pyvista (works in console and notebooks)
  - `export_geometry`: Save the mesh as `vts` or `vtp`.
- Added an iterator to the domain class so that we can loop over all domains
- Added 3D views for tutorial 6
- Added the neccesary dependencies.

You can now view the domain with:

```python
from struphy import domains
domain = domains.Tokamak()
domain.show_3d()
```

image

To display all domains (except for the Spline ones, which don't work with default params) you can do this:

```python
from struphy.geometry.base import Domain

for domain_cls in Domain:
    if "Spline" in domain_cls.__name__:
        continue
    print(domain_cls.__name__)
    domain = domain_cls()
    domain.show_3d()
```


max-models:
Before merging, try out the 3D figures in the tutorials, they work on my Mac, but you should check it too because I had some trouble getting it to work in the CI @spossann.

.github/workflows/submod-struphy-tutorials.yml
```diff
@@ -33,6 +33,10 @@ jobs:
         if: env.SUBMOD_CHANGED == 'true'
         uses: ./.github/actions/install/ubuntu-latest
       
+      - name: Install OSMesa/EGL
+        if: env.SUBMOD_CHANGED == 'true'
+        run: sudo apt-get install -y libosmesa6 libosmesa6-dev libegl-mesa0
```
spossann, line 38: These libraries are only needed in the CI I guess?

pyproject.toml
```diff
@@ -34,8 +34,13 @@ dependencies = [
...
     'argcomplete',
+    'ipywidgets',
     'plotly',
-    "pyvista",
+    'pyvista',
+    'trame',
+    'trame-vtk',
+    'trame-vuetify',
+    'vtk',
     'pytest',
```
spossann, line 45: I had to install `nest_asyncio2` for the interactive display in notebooks. I think we shoud add it to the dependencies.

spossann (dismissed):
The interactive view is super nice!

.github/workflows/submod-struphy-tutorials.yml
max-models: I think so, the issue was that when converting the notebooks in the CI, we have to turn off the interactive view.

pyproject.toml
max-models: Ok! I will add it.

max-models: Fixed in [2ffd6c7](https://github.com/struphy-hub/struphy/pull/195/commits/2ffd6c71b0c33b16d2a4273227ce74ca4d12803e)

spossann (approved):

PR #205, opened 2026-03-13
202 toygyrokinetic model

### LINKED ISSUE #202: toyGyrokinetic model
Create a new model `ToyGyrokinetic` to implement the **dicotron instability**

The new model is based on the `DriftKineticElectrostaticAdiabatic` model by removing the `PushGuidingCenterParallel` propagator.

Changes to the new model:
1. The effective electric field

$$
\mathbf{E}^* = - \nabla \phi - \epsilon \mu_p \nabla |\mathbf{B}|
$$

is replaced with

$$
\mathbf{E} = -\nabla \phi
$$

This is done in the parameter file by setting `b_tilde` of the `PushGuidingCenterBxEstar` to None and using Homogenslab fluid equilibrium.

2. Remove PushGuidingCenterParallel propagator

P-Attapon:
Simulate the Diocotron instability

**Changes to the source code**

1. New model `ToyGyrokinetic`
2. `show_accumulated_spline_field` used to debug charge accumulation can now show accumulation result in 2D
3. `ModesCos` perturbation is adjusted to allow perturbation within a subdomain.

For the subdomain perturbation, the method currently takes in the argument in logical space. It can be improve to accept the argument in coordinate of the domain and generalizes to other type of perturbations.

**Changes to the examples test cases**
1. Diocotron instability
1.1 Add a parameter file for Diocotron instability
1.2 Add a post-processing file for the Diocotron instability
2. Weibel instability
2.1 Edit the Weibel instability test case to show the Gauss error
2.2 post-process from `path_out` instead of `params.sim`


src/struphy/models/toy_drift.py
```diff
@@ -0,0 +1,208 @@
...
+            "H1",
+            Pyccelkernel(accum_kernels_gc.gc_density_0form),
+            Propagator.mass_ops,
+            Propagator.domain.args_domain,
+        )
+
+        rho = charge_accum
+
+        # sanity check: compute FE coeffs of density
+        # rho()
```
P-Attapon, line 151: Here I have to call the accumulator before showing the spline field otherwise it's zero in the whole space

src/struphy/pic/accumulation/particles_to_grid.py
```diff
@@ -644,18 +648,39 @@ def show_accumulated_spline_field(self, mass_ops, eta_direction=0):
...
+        if len(plt_axis) == 1:
+            plt.plot(eta, field_value)
+
+            plt.xlabel(rf"$\eta_{plt_axis[0] + 1}$")
+            plt.ylabel("field amplitude")
+
+        # Two-dimensional case
+        elif len(plt_axis) == 2:
+            Eta1, Eta2 = xp.meshgrid(eta, eta)
+            pcm = plt.pcolor(Eta1, Eta2, field_value)
```
P-Attapon, line 676: Sorry again for this confusion of mine. But should the field_value here be transposed?

P-Attapon:
@spossann I've implemented the visualization of charge accumulation from poisson equation (`show_accumulated_spline_field`). 

Below is the plot from the Diocotron instability. The initial condition is assigned such that all charges should be located within `r_min, r_max = 4.0, 5.0`. This corresponds to `eta_min, eta_max = 0.33, 0.44` in logical space. 

There are two comments in the files that I would ask you to please confirm. let me know your thoughts :)

**Charge accumulation from Poisson solve:**
poisson_solve_accumulation

src/struphy/models/toy_drift.py
```diff
@@ -0,0 +1,208 @@
...
+
+    1. :class:`~struphy.propagators.propagators_fields.ImplicitDiffusion`
+    2. :class:`~struphy.propagators.propagators_markers.PushGuidingCenterBxEstar`
+
+    :ref:`Model info <add_model>`:
+    """
+
+    @classmethod
+    def model_type(cls) -> LiteralOptions.ModelTypes:
+        return "Kinetic"
```
spossann:
```suggestion
        return "Toy"
```

src/struphy/models/toy_drift.py
```diff
@@ -0,0 +1,208 @@
...
+    class KineticIons(ParticleSpecies):
+        def __init__(self):
+            self.var = PICVariable(space="Particles5D")
+            self.init_variables()
+
+    ## propagators
+
+    class Propagators:
+        def __init__(self):
+            self.gc_poisson = propagators_fields.ImplicitDiffusion()
```
spossann:
```suggestion
            self.gc_poisson = propagators_fields.Poisson()
```

src/struphy/models/toy_drift.py
```diff
@@ -0,0 +1,208 @@
...
+        # rho()
+        # rho.show_accumulated_spline_field(Propagator.mass_ops, eta_direction=(True,True,False))
+
+        self.propagators.gc_poisson.options.sigma_1 = 1e-8
+        self.propagators.gc_poisson.options.sigma_2 = 0.0
+        self.propagators.gc_poisson.options.sigma_3 = 1.0
+        self.propagators.gc_poisson.options.stab_mat = "M0ad"
+        self.propagators.gc_poisson.options.diffusion_mat = "M1"
+        self.propagators.gc_poisson.options.rho = rho
+        self.propagators.gc_poisson.allocate()
```
spossann: If we take the `Poisson` propagator, then these options will change.

src/struphy/models/toy_drift.py
```diff
@@ -0,0 +1,208 @@
...
+
+        # energy from polarization
+        e1 = Propagator.derham.grad.dot(-phi, out=self._e_field)
+        en_phi1 = 0.5 * Propagator.mass_ops.M1gyro.dot_inner(e1, e1)
+
+        # for Landau damping test
+        # en_phi = 0.
+
+        # mu_p * |B0(eta_p)|
+        particles.save_magnetic_background_energy()
```
spossann: Not needed.

src/struphy/models/toy_drift.py
```diff
@@ -0,0 +1,208 @@
...
+
+        # 1/N sum_p (w_p v_p^2/2 + mu_p |B0|_p)
+        self._tmp3[0] = (
+            1
+            / particles.Np
+            * xp.sum(
+                particles.weights * particles.velocities[:, 0] ** 2 / 2.0 + particles.markers_wo_holes_and_ghost[:, 8],
```
spossann:
```suggestion
                particles.weights * particles.velocities[:, 0] ** 2 / 2.0
```

src/struphy/pic/accumulation/particles_to_grid.py
```diff
@@ -644,18 +648,39 @@ def show_accumulated_spline_field(self, mass_ops, eta_direction=0):
...
+        # One-dimensional case
+        if len(plt_axis) == 1:
+            plt.plot(eta, field_value)
+
+            plt.xlabel(rf"$\eta_{plt_axis[0] + 1}$")
+            plt.ylabel("field amplitude")
+
+        # Two-dimensional case
+        elif len(plt_axis) == 2:
+            Eta1, Eta2 = xp.meshgrid(eta, eta)
```
spossann:
```suggestion
            Eta1, Eta2 = xp.meshgrid(eta, eta, indexing='ij')
```

src/struphy/pic/accumulation/particles_to_grid.py
spossann: Try the more natural meshgrid indexing above.

spossann (commented):
This looks good already. We need to figure out the transpose and what happened to the boundary conditions in the plot (there is a step at the y-boundaries).

P-Attapon:
@spossann Here, are the list of updates I've made to the code

1. A small remark that, in the calculation of energy in particles, I've removed the contribution from the adiabatic electrons. From my understanding it comes from the kinetic parallel to the magnetic field which we don't use in the model.
```
# energy from adiabatic electrons
en_phi = 0.5 / epsilon**2 * Propagator.mass_ops.M0ad.dot_inner(phi, phi)
```

2. I've changed the indexing of the meshgrid in `show_accumulated_spline_field` to 'ij' as suggested. This is the result:
poisson_solve_fixIndexing

Could the step at the boundary be caused by our setup of dirichlet boundary condition?

```
# Derham options
derham_opts = DerhamOptions(
    p=(3,3,1), 

    # impose dirichlet boundary conditions at r_min and r_max
    spl_kind=(False,True,True), 
    dirichlet_bc=(
        (True, True),
        (False, False),
        (False, False),
    ))
```

3. I replaced the `ImplicitDiffusion` propagator with the `Poisson`. Here are the results from the potential field solved from the `ImplicitDiffsion` with `M1` diffusion matrix and the `Poisson`. Now, the field no longer explodes, but the energy in $\phi$ and particles remain constant.

**ImplicitDiffusion with M1 diffusion matrix**:
potentialEvolution

**Poisson:**
potentialEvolution

spossann:
Ok good!
You are right that the Dirichlet boundary conditions which we set for the derham complex cause the step at the boundary. That's ok, since this is only a test view. But why is the density negative -1 in most of the domain?

P-Attapon:
> Ok good! You are right that the Dirichlet boundary conditions which we set for the derham complex cause the step at the boundary. That's ok, since this is only a test view. But why is the density negative -1 in most of the domain?

I ran the simulation again in the full-f case with `control_variate = False`. I think that the -1 is related to the background distribution `maxwellians.GyroMaxwellian2D(n=(1.0, None), equil=equil)`
poisson_solve_CVF

spossann:
> > Ok good! You are right that the Dirichlet boundary conditions which we set for the derham complex cause the step at the boundary. That's ok, since this is only a test view. But why is the density negative -1 in most of the domain?
> 
> I ran the simulation again in the full-f case with `control_variate = False`. I think that the -1 is related to the background distribution `maxwellians.GyroMaxwellian2D(n=(1.0, None), equil=equil)` poisson_solve_CVF

Ok, I think this is a very important point for the Poisson solve. Indeed, we want the density rho on the right-hand side to be the full-f density, thus accumulated with the constant weights w0, not with the delta-f weights obtained from `update_weights`. Please check.

src/struphy/models/toy_drift.py
```diff
@@ -0,0 +1,205 @@
...
+        :meta private:
+        """
+        self._tmp3 = xp.empty(1, dtype=float)
+        self._e_field = Propagator.derham.Vh["1"].zeros()
+
+        assert self.kinetic_ions.charge_number > 0, "Model written only for positive ions."
+
+        # Poisson right-hand side
+        particles = self.kinetic_ions.var.particles
+        particles.weights = particles.weights_at_t0.copy()
```
P-Attapon, line 135: @spossann

Here, is my current solution. update_weights is applied within the initialize_weights function. I think that's the reason why full-f was not previously used in the accumulation.

For now I set the full-f here manually and call update_weights later if control_variate=True.

Image

spossann: Ok, does the solution of the Poisson equation now look correct?  It should see the cosine in eta2 somehow

P-Attapon: > Ok, does the solution of the Poisson equation now look correct? It should see the cosine in eta2 somehow

potentialEvolution
We see that in other time step, but the initial is still uniform

P-Attapon: > Ok, does the solution of the Poisson equation now look correct? It should see the cosine in eta2 somehow

Here, I put in a higher value of perturbation (amps = 20) to see the cosine better.

poisson_solve_amp20
potentialEvolution

spossann: Nice - thanks! It seems the simulation should run correctly then. You can try to run with a coarse resolution (16, 32, 1) for many time steps I guess.

P-Attapon: > Nice - thanks! It seems the simulation should run correctly then. You can try to run with a coarse resolution (16, 32, 1) for many time steps I guess.

Right now there is one problem that nothing seems to be moving within the simulation. I am not really sure about the cause.

growth_rate

e1_e2_density_delta_f_binned_phaseSpace

potentialEvolution

P-Attapon:
@spossann 

I think that I've successfully reproduced the diocotron instability. Here is the result for the non-linear case with perturbation amplitude of 0.5

e1_e2_density_f_binned_phaseSpace
 
Here is for the linear case (amplitude = 1e-6)
e1_e2_density_f_binned_phaseSpace

spossann:
Very good!

P-Attapon:
I was cleaning up the weibel instability, and may discovered a possible bug in the post-processing of data. I have not thoroughly checked why, but I received different results  when post process with `PostProcessor(sim=<sim>)` and `PostProcessor(path_out=<file>)` 

This is the result when process with `path_out` (expected result)
growth
init_EM

This is the result when process with `sim`
growth1
init_EM1

max-models:
> @spossann
> 
> I think that I've successfully reproduced the diocotron instability. Here is the result for the non-linear case with perturbation amplitude of 0.5
> 
> e1_e2_density_f_binned_phaseSpace
> Here is for the linear case (amplitude = 1e-6) e1_e2_density_f_binned_phaseSpace

Very cool!!

spossann (approved):

spossann:
Hi @P-Attapon the bug in post-processing from the Simulation object was fixed in [8e044cd](https://github.com/struphy-hub/struphy/pull/205/commits/8e044cd00dbab6d1891ec06d55199b27d749bc9c)

PR #199, opened 2026-03-08
Use logging instead of print

max-models:
Based on this tutorial: https://www.youtube.com/watch?v=9L77QExPmI0

- [x] ~~Add logging config~~ (Now it's just a dict)
- [x] Allow for overriding the logging level as an arg into Simulation
- [x] Add a rank filter so only rank == 0 actually logs
- [x] Replace all prints with logging

Here are the logging levels:

image

I would suggest that the cases where we have

```python
if verbose:
    print(message)
```

could be replaced by `logger.debug(message)`.

Then we can set the default logging to INFO and if you wish to change it, you just set it to DEBUG and then all the verbose logging messages will be show.

Also, another nice thing is that we can add this rank filter so that instead of doing:

```python
if self.rank == 0:
    print(message)
```

we can just do `logger.debug(message)` and by default only rank == 0 logs. We can also add more handlers to which this filter does not apply, which might be the case for certain types of debugging, where you might want to log from every rank.

PR #219, opened 2026-04-02
Set logger in struphy __init__.py

spossann:
This PR goes into `use-logging-instead-of-print`, see https://github.com/struphy-hub/struphy/pull/199

I read that the logger should be set up at the beginning of the entry point of the program. Since we do not have a clear entry point, the best thing is to maybe setup in the struphy `__init__.py`.

* move logger setup to struphy `__init__.py`
* do not use .json file to avoid importing struphy for the path
* use function `set_logger_level` to set the level of all handlers 

The function `set_logger_level` can be tested with `python src/struphy/utils/utils.py`.

I set the default level to `logging.WARNING`.


src/struphy/__init__.py
```diff
@@ -1,3 +1,105 @@
...
+            queue_handler = handler
+            break
+    if queue_handler is not None:
+        queue_handler.listener.start()
+        atexit.register(queue_handler.listener.stop)
+
+# Default logging setup
+logger = logging.getLogger("struphy")
+setup_logging(logging_level=logging.WARNING)
+logger.info("Logging setup complete.")
```
max-models, line 100: If we now run the logging setup in the `__init__.py`, I assume it's run on the first import of struphy, can I then change the logging like this? Or how do you think would be the best way to change the logging level?

```python
import logging
import struphy
struphy.setup_logging(logging_level=logging.INFO)
```

spossann: Yes indeed, I would do 
```
import logging
from struphy import set_logging_level

set_logging_level(logging.DEBUG)
```
as in the test here: [set_logging_level(logging.DEBUG)](https://github.com/struphy-hub/struphy/pull/219/changes#diff-a0213cb1b5f86b154d266ab93069aeba621a3a564471c3c65cc2f077c1744900R204)

max-models: Ok, I think that looks good! I'll merge this into my branch then.

PR #220, opened 2026-04-09
New classes for scalar quantities tracked during sim

spossann:
Up to now, the handling of scalars in `StruphyModel` was via `scalar_quantities: dict`, which still stems from the "old" struphy. This is now replaced by `scalars : Scalars or None` which is a new class `Scalars` containing objects of the new type `Scalar`, which is an abstract base class with several subclasses implemented in `models/scalars.py`.

The following has been removed from `StruphyModel`:
* `add_scalar` (method)
*  `update_scalar` (method)
* `update_scalar_quantities` (abstract method)
* `scalar_quantities: dict` (property)

All this has been replaced by the property `scalars`.


spossann:
@max-models I would like to get your opinion on this before I go ahead and adapt all model accordingly. The models `Maxwell` and `VlasovAmpereOneSpecies` have already been adapted and can be checked as examples. Thx!

src/struphy/models/scalars.py
```diff
@@ -0,0 +1,261 @@
...
+        """Update the scalar quantity by performing local update and then summing over MPI processes."""
+        if not self.uptodate:
+            self._local_update()
+            self._mpi_sum()
+            self.uptodate = True
+
+    def __add__(self, other):
+        return SumOfScalars(self, other)
+
+class SumOfScalars(Scalar):
```
max-models, line 52: Should this class not have an `__add__` method too? For summing three scalars for example

max-models: Or maybe tit already works by inheritance

spossann: Yes, it should have the `__add__` by inheritance.

max-models:
> @max-models I would like to get your opinion on this before I go ahead and adapt all model accordingly. The models `Maxwell` and `VlasovAmpereOneSpecies` have already been adapted and can be checked as examples. Thx!

I think this is a good idea! I don't really have any opinions on the exact implementation, what you did looks reasonable. So if it works, then it's good enough for me!

max-models (approved):

PR #222, opened 2026-04-09
Compute equation parameters in model __init__

### LINKED ISSUE #221: Access equation parameters before runtime
The equation parameters (`Species.equation_parameters`, like `alpha`, `epsilon` etc) must be set at runtime to normalize the model:

 https://github.com/struphy-hub/struphy/blob/9a4495f292848825e6d04e8e1de837ae3648a9a9/src/struphy/simulation/sim.py#L275

 This is because they require information from the launch file like charge- and mass number of the species. However, the equation parameters are often needed in the definition of propagators or scalars as normalizing coefficients. It would be nice to address them before "runtime".

**Proposed solution:**

We compute the equation parameters in the model __init__. This requires passing the species charge- and mass numbers when instantiating the model.

spossann:

Closes #221 

**API changes:**

* `base_units` is removed from the `StruphyModel` constructor.
* All equation parameters that can be seen in the model docstring can be passed to the model constructor. 

**Core changes:**

The method `Species.set_species_properties` is removed and replaced by augmenting `Species.init_variables`. This means that the species properties are available after instantiation of the species, in the model __init__. Moreover, we now pass the arguments of `set_species_properties`, which was previously called in the launch file for instance here: https://github.com/struphy-hub/struphy/blob/9a4495f292848825e6d04e8e1de837ae3648a9a9/examples/VlasovAmpereOneSpecies/two_stream/params_two_stream.py#L52 directly to the model constructor. 

In addition, in order to compute the normalization parameters of the PDE (alpha, epsilon, etc.), we need to pass the `BaseUnits` to the model constructor. In this way everything can be computed with `StruphyModel.setup_equation_params`, which is called in the model __init__ before Propagators or Scalars are instantiated.

For example, a typical model creation would now look like this:
```
# Units
base_units = BaseUnits()

# Model instance
model = VlasovAmpereOneSpecies(base_units=base_units, alpha=1.0, epsilon=-1.0, with_B0=False)
```
Note that the base units have to be created before the model instance; this has been changed in the default parameter file creation. Also, there are defaults such that
```
# Model instance
model = VlasovAmpereOneSpecies()
```
still works.


spossann:
@max-models this is another PR where I would like your opinion before I go ahead with changing all the models. You can look at `VlasovAmpereOneSpecies` for an example of how it looks like. Thx again!

max-models (approved):

PR #237, opened 2026-05-05
Update ci feectools install

spossann:
Update CI:

- [x] remove some triggers on push to `devel`
- [x] re-install `feectools` only if there are submodule changes detected 
- [x] use upper bounds in `dependencies`, `phys` and `mpi` in the `devel` branch. These ar checked for major and minor release updates on every PR into devel. The policy is summarized in `CONTRIBUTING.m`:

Pull requests into **devel** are checked for stale dependency upper bounds in `pyproject.toml`.

The policy is intentionally narrow:

1. Only `project.dependencies` are checked.
2. From `project.optional-dependencies`, only `phys` and `mpi` is checked.
3. Patch-only updates are ignored.
4. A pull request fails when a newer **major** or **minor** stable release exists on PyPI beyond the declared upper bound.

When the check fails, the CI summary prints local remediation commands. In short, run the checker locally, run `python utils/update_dependency_bounds.py` on that report, and commit the updated `pyproject.toml`.

Try with `python utils/check_dependency_bounds.py`.

One can test this by lowering a version number in pyproject.toml and run the above command again. In the pipeline, then one will get instructions of how to proceed.


max-models:
> A pull request fails when a newer **major** or **minor** stable release exists on PyPI beyond the declared upper bound.

What happens if we actually require the version to be lower than the latest version on PyPI?

spossann:
> > A pull request fails when a newer **major** or **minor** stable release exists on PyPI beyond the declared upper bound.
> 
> What happens if we actually require the version to be lower than the latest version on PyPI?

Good point. At the moment it is planned to not allow this, which forces us to patch immediately when we are not compatible with the newest version of a dependency.

Should we add the handling of exceptions? 

Btw I re-added `pip install upgrade-strategy eager` so to test the newest versions always.

max-models:
> > > A pull request fails when a newer **major** or **minor** stable release exists on PyPI beyond the declared upper bound.
> > 
> > 
> > What happens if we actually require the version to be lower than the latest version on PyPI?
> 
> Good point. At the moment it is planned to not allow this, which forces us to patch immediately when we are not compatible with the newest version of a dependency.
> 
> Should we add the handling of exceptions?
> 
> Btw I re-added `pip install upgrade-strategy eager` so to test the newest versions always.

No need to handle an exception that doesn't apply yet. We can worry about this once it occurs. The only dependency which previously had an upper bound was `jax-finufft<=1.2.0`, but it has been removed in this PR anyway.

max-models (approved):

PR #243, opened 2026-05-11
CI memory shortage

### LINKED ISSUE #242: Github CI runs out of memory
This happens since May 7: https://github.com/struphy-hub/struphy/actions/workflows/test-PR-unit-mpi.yml

spossann:

Closes #242 

I guess this is because Github changed memory limits? Is there any source for this?

The reason is most like the new garbage collector introduced in Python 3.14. It can cause problems in large pytest runs, since packages like numpy, pandas etc have not yet adopted it. Actually, the new GC has been rolled back in Python 3.14.5, see https://www.igorslab.de/en/python-3-14-5-released-maintenance-update-old-garbage-collector/

**Implemented solution:**

Pin Python to 3.13 in the docker image `ubunte-latest-with-struphy`. The other image`ubuntu-with-reqs` still has the newest Python version and can be used for scheduled testing.

**Other changes:**

Implemented sharding of unit tests (4 shards) to have quicker concurrent workflows.


spossann:
Hi @max-models - this is now indeed ready for review! The reason for the memory overflow was a change in the garbage collection in Python 3.14 (see description above).

max-models (approved):
This solution looks a bit like a hack, but maybe it's the best way to solve it for now. :)

PR #244, opened 2026-05-13
Add new `CurlCurlSolve()` to propagators

Monoclod:
Add a new propagator called `CurlCurlSolve()` to find weak solutions to the curl-curl problem.

Find $\mathbf{E} \in H(\textnormal{curl},\Omega)$ such that 

$$
\int_\Omega (\nabla \times \mathbf F) \cdot (\nabla \times \mathbf E)\ \textrm{d} \mathbf x - \sigma \int_\Omega \mathbf F \cdot \mathbf E\ \textrm d \mathbf x = \sum_i \int_\Omega \mathbf F \cdot \mathbf J _i\ \textrm d \mathbf x \qquad \forall \mathbf F \in H(\textnormal{curl})\,
$$

where $\mathbf J _i:\Omega \to \mathbb R^3$ is a real-valued vector field and $\sigma \in \mathbb R$ is a scalar.
Boundary terms from integration by parts are assumed to vanish.

The equation is discretized as

$$
\left( \mathbb C^\top \cdot \mathbb M^2 \cdot \mathbb C - \sigma \mathbb M^1 \right)\cdot \boldsymbol e^{n+1} =\sum_i \mathbb P^1 \cdot \boldsymbol J _i \,
$$

where $M^1$ and $M^2$ are objects of the `struphy.feec.mass.WeightedMassOperators` class, and $\mathbb P^1$ is the projector into the space $V^1_h$.

The method does not converge properly for sufficiently high spline degrees and sufficiently fine mesh grids due to the inherent ill-conditionedness of the curl-curl problem; this should be solved by proper initial conditioning.

### DIFF
--- a/src/struphy/propagators/curl_curl_solve.py
+++ b/src/struphy/propagators/curl_curl_solve.py
@@ -1,0 +1,323 @@
+import logging
+from dataclasses import dataclass
+from typing import Callable, Literal
+
+import cunumpy as xp
+from feectools.ddm.mpi import mpi as MPI
+from feectools.linalg.basic import IdentityOperator
+from feectools.linalg.solvers import inverse
+from feectools.linalg.stencil import StencilVector
+from line_profiler import profile
+
+from struphy.feec.mass import L2Projector, WeightedMassOperator
+from struphy.io.options import LiteralOptions
+from struphy.linear_algebra.solver import SolverParameters
+from struphy.models.variables import FEECVariable
+from struphy.pic.accumulation.particles_to_grid import AccumulatorVector
+from struphy.pic.base import Particles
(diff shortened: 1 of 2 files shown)


spossann:
@Monoclod well done! Please add some tests that prove that the new propagator works. For the convergence of higher degrees, I suggest you take a function with more variations, so that the error is larger, and set `tol=1e-7` in the solver.

Monoclod:
Added a test for the propagator (and fixed a few comments). They seem to converge fine even for tolerances of 1e-10, oddly better than my previous attempt

spossann:
@Monoclod can you merge with devel to see if the tests pass?

spossann:
will be merged from another branch.

PR #246, opened 2026-05-21
Implementation of average linear operator

emilegrivet:
Addition of a new linear operator for averaging a StencilVector along a direction (1D averaging).

The operator is matrix free.

A test for this new operator is available in src/struphy/feec/tests/test_mass_matrices.py


spossann (approved):
Very nice @emilegrivet ! I added ssg to the docstring.

PR #248, opened 2026-05-28
new feectools commit for repairing inverse method

spossann:


max-models:
Before merging this, we should merge the commit from devel-tiny in feectools.

spossann:
> Before merging this, we should merge the commit from devel-tiny in feectools.

We need a strategy for `feectools`:

At the moment, a push to `devel-tiny` triggers a release on PyPI, and thus any PR into `devel-tiny` needs a new version number. There is no development branch. For the existing `struphy` release this is fine, because we have an upper bound on the `feectools` dependency, which is updated on release.

In a struphy branch, the submodule `feectools` can in principle be pinned to any commit. In `devel` we should maybe pin it to `devel-tiny`.

There is a list of PRs into `devel-tiny`, which we can work through on Monday.

spossann:
Handled in #332

PR #251, opened 2026-06-02
Modification of diocotron example

emilegrivet:
Small modifications on pproc.py in order to compare more easily simulations performed in the diocotron instability example folder.


spossann:
Has been integrated with #260

PR #260, opened 2026-06-10
Cyclone test case

emilegrivet:
Added a different safety factor profile for `adhocTorus` (parabolic with a linear term).

Added an `r_min` parameter to the `Tokamak` domain.

Added an implicit diffusion propagator for the Poisson solve with adiabatic electrons, named `PoissonAdiabaticGyrokinetic`. Both `ImplicitDiffusion` and both poisson propagators can now also take `Particles` and `PICVariable` objects as sources.

Added two new examples based on the `DriftKineticElectrostaticAdiabatic` model: itg_cylinder and cyclone. The cyclone test case still exhibits some issues due to noise developing during the simulation.

Modified the assembly of weighted mass operators to ensure consistent block assembly across different MPI processes.


src/struphy/feec/mass.py
```diff
@@ -1532,46 +1530,55 @@ def __init__(
...
-                                            backend=PSYDAC_BACKEND_GPYCCEL,
-                                            precompiled=True,
-                                        ),
-                                    ]
-                                self._weights[-1] += [weights_info[a][b]]
+                            assert mat_w.shape == tuple([pt.size for pt in pts])
+                            local_nonzero = xp.array(bool(xp.any(xp.abs(mat_w) > 1e-14)), dtype=bool)
+
+                        if self.derham.comm is not None:
+                            self.derham.comm.Allreduce(MPI.IN_PLACE, local_nonzero, op=MPI.LOR)
```
spossann, line 1646: This is an important bug fix that needs to be explained in the PR description. Maybe also add a comment to these lines what is checked here. I assume that `None` for a block is only acceptable if that block is zero on all ranks. What happens if a block is `None` only on some ranks?

src/struphy/geometry/domains.py
```diff
@@ -41,6 +41,8 @@ class Tokamak(PoloidalSplineTorus):
         Parametrization of radial flux coordinate :math:`\eta_1=\psi_{\mathrm{norm}}^p`, where :math:`\psi_{\mathrm{norm}}` is the normalized poloidal flux (default: 0.75).
     psi_shifts : tuple[float]
         Start and end shifts of polidal flux in % --> cuts away regions at the axis and edge (default: [2., 2.])
+    r_min : float
+        Inner radius of poloidal section (optionnal, default: 0.0). If >0.0, then r_0 = r_min.
```
spossann: typo in `optionall`

src/struphy/geometry/domains.py
```diff
@@ -94,9 +100,30 @@ def __init__(
         # get control points via field tracing between fluxes [psi_s, psi_e]
         psi0, psi1 = equilibrium.psi_range[0], equilibrium.psi_range[1]
 
-        psi_s = psi0 + psi_shifts[0] * 0.01 * (psi1 - psi0)
+        assert r_min >= 0.0, f"Inner radius must be non-negative, got {r_min = }."
+
+        if r_min == 0.0:
+            # Default behaviour: keep exactly the historical psi_shifts logic.
+            psi_s = psi0 + psi_shifts[0] * 0.01 * (psi1 - psi0)
```
spossann, line 116: I think `psi_shifts` already does the trick, is the new code redundant? You can test this in the tutorial `tutorial_mhd_equilibira.ipynb`.

spossann (commented):
Looks good! I will leave further comments later.

src/struphy/geometry/domains.py
emilegrivet: If r_min is left at 0.0, then psi_shifts provides the inner (and outer) radius in terms of the flux \psi. If r_min > 0.0, we do not use psi_shifts[0] given in the parameters; instead, we calculate an approximation of the flux coordinate corresponding to the inner radius r_min. This allows us to provide a more intuitive inner boundary. In this case, psi_shifts[1] is still used for outer boundary.

examples/DriftKineticElectrostaticAdiabatic/itg_cylindre/pproc_drift_kinetic.py
```diff
@@ -472,13 +472,6 @@ def update(_):
 # Main
 # ============================================================
 def main():
-    '''if len(sys.argv) > 1 and __name__ == "__main__":
```
spossann: Is this missing now in the file? Should it be added at the end?

spossann: Oh I just saw it is already there.

src/struphy/feec/mass.py
spossann: adressed in [096613e](https://github.com/struphy-hub/struphy/pull/260/commits/096613ebce33a67b672b5613b31105af74c71a64)

spossann:
@emilegrivet there is sill a barrier that is not reached it seems.

spossann (approved):

PR #263, opened 2026-07-03
Tok cluster setup

spossann:
* Set a local path for the logging file instead of global `/tmp/strupy.log`. The global path would lead to errors for users other than the creator of the path.
* Use print commands in compile statements so that you always see what is happening (for instance with `struphy compile --status`)
* Add some install instructions for the TOK cluster to the doc.


src/struphy/__init__.py
```diff
@@ -37,7 +37,7 @@ def filter(self, record):
             "class": "logging.handlers.RotatingFileHandler",
             "level": "WARNING",
             "formatter": "detailed",
-            "filename": "/tmp/struphy.log",
+            "filename": "tmp/struphy.log",
```
max-models: Maybe we don't even need the `tmp/` here, why not just save the log to `struphy.log`?

spossann: changed in [f729d50](https://github.com/struphy-hub/struphy/pull/263/commits/f729d508e8af8da01c49ad3a459e68c1dfdb3c78)

max-models (approved):

PR #266, opened 2026-07-04
Automated scaling test

max-models:
The profiling data is published [here](https://struphy-hub.github.io/profiling-data/)

## Definition of a **case**

- Specific compiled version of struphy (pyccel language etc)
- Specific setup of python dependencies
- Specific physics model
- Specific modules loaded, environment variables, ...
- Parametric variations can occur, both on discretization, hardware etc, but NOT on the previously mentioned points.

## How is the metadata stored?

- Store each *case* (for example, a strong scaling testcase) in one folder.
- Store one `case_metadata.json` with global case metadata
- Store one `run_metadata.json` for each `sim.run()` event, this contains specific metadata for that ran (number of ranks, ppc, Nel, ...). This means the `sim.env`, `sim.time_opts`, `sim.grid`, `sim.derham_opts`. Once https://github.com/struphy-hub/struphy/issues/273 is done, we should just use this json file.

## What to store in the  `case_metadata.json` metadata files

**General information**
- Time/date
- User
- slurm script
- slurm variables
- Name of the test case
- Description of the test case
- Physics problem
- Struphy model used

**Hardware information**
- cluster name
- machine information, chip name, etc
- nodes

**Software information**
- parameter file
- python environment (pip freeze)
- environment variables
- modules
- struphy commit
- pyccel language
- pyccel compiler family
- pyccel openMP
- pyccel metadata
- struphy compile command output

## Usage

```
python profiling/submit_diocotron_job.py
```

Currently, only a Pitagora job for a simple test case is added, but I will make it more general.


profiling/examples/ToyGyrokinetic/diocotron_instability/params_diocotron.py
spossann, line 1: Is this a copy of the file in `examples/`? If so, why not just take that file? Also future cases could just be based on examples.

hello.sh
spossann, line 1: Delete?

spossann (dismissed):

profiling/examples/ToyGyrokinetic/diocotron_instability/params_diocotron.py
max-models: It's almost a copy, I changed `boxes_per_dim` to allow for running it with more MPI ranks. I was thinking "profiling examples" would be a bit separate from the usual examples since they may not be runnable on the github runners if many nodes are needed.

hello.sh
max-models: Done in [1107a18](https://github.com/struphy-hub/struphy/pull/266/commits/1107a18866a1898c7c4ad58b65c39bc3babd16dc)

spossann (approved):

PR #276, opened 2026-07-13
added new example for coldplasmavlasov, removed examples from gitignore

anishojhaofficial:
As the test case in the Master Thesis 2018


spossann:
Hello!

PR #283, opened 2026-07-20
Added workflow to push docs to struphy-hub/docs-preview

### LINKED ISSUE #255: Publish preview docs in a preview repository
Reviewing documentation changes currently requires checking out the branch and building the docs locally.

Set up automated preview deployments for pull requests so reviewers can view the rendered documentation before changes are merged. On each PR update, the documentation should be built and published to a GitHub Pages preview site under a PR-specific URL.

One option would be to use to use commercial software, such as Vercel, Netlify, Cloudflare, which handle some of this automatically. But I think it would be simpler (and free) to just do it with a separate git repo.

What I have in mind is the following:

- Create a separate GitHub Pages repository for documentation previews.
- On each PR update, build the docs in the `struphy` repository.
- Push the generated static files to the preview repository under a PR-specific folder (e.g. `pr-123/`).
- GitHub Pages publishes the preview automatically. Each preview can be found under `https://struphy-hub.github.io/docs-preview/pr-123/`
- Remove the preview folder when the PR is closed with a second workflow.

max-models:

Closes #255

### DIFF
--- a/.github/workflows/docs-preview.yml
+++ b/.github/workflows/docs-preview.yml
@@ -117,7 +117,7 @@ jobs:
             const prNumber = context.payload.pull_request.number;
             const url = `https://struphy-hub.github.io/docs-preview/pr-${prNumber}/`;
             const marker = '<!-- docs-preview-link -->';
-            const body = `${marker}\n📖 Docs preview: ${url}\n\n_Updated automatically on each push to this PR._`;
+            const body = `${marker}\n📖 Docs preview: ${url}`;
 
             const { data: comments } = await github.rest.issues.listComments({
               owner: context.repo.owner,
@@ -126,14 +126,7 @@ jobs:
             });
             const existing = comments.find(c => c.body.includes(marker));
 
-            if (existing) {
-              await github.rest.issues.updateComment({
-                owner: context.repo.owner,
-                repo: context.repo.repo,
-                comment_id: existing.id,
-                body,
-              });
-            } else {
+            if (!existing) {
               await github.rest.issues.createComment({
                 owner: context.repo.owner,
                 repo: context.repo.repo,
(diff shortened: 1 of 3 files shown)


doc/index.rst
```diff
@@ -8,7 +8,7 @@ Welcome to
 
 |
 
-This website is your starting point for solving PDEs with Struphy.
+This website is your starting point for solving PDEs with Struphy!
```
spossann, line 11: LOL

.github/workflows/docs-preview.yml
```diff
@@ -0,0 +1,143 @@
+name: Deploy PR docs preview
+
+on:
+  pull_request:
+    branches:
+      - devel
+    types: [opened, synchronize, reopened]
```
spossann, line 7: This workflow run will be very long: install and compile struphy and build the docs. I suggest to add a `workflow_dispatch:` and not run it on every PR into devel.

spossann (commented):
We need this!

.github/workflows/docs-preview.yml
max-models: I was thinking we could maybe activate the workflow manually or with a specific tag in the commit message like `[preview docs]`.

max-models: Done in [dcf756f](https://github.com/struphy-hub/struphy/pull/283/commits/dcf756f9c5f5fde3f39c6a855edac61c69532ee8)

Testing `[preview docs]` in https://github.com/struphy-hub/struphy/actions/runs/29828069618/job/88625930549?pr=283

spossann: It seems this workflow runs only when manually triggered through dispatch OR if preview or docs figures in the commit message, right?

max-models: Yes exactly, manually triggered through dispatch OR the string `[preview docs]` should be in the commit message

spossann (approved):

Issue #172, opened 2026-02-06
Add a `struphy build-init-files` or something similar.


max-models:
Because this is not "formatting"

PR #286, opened 2026-07-22
Resolve 171 init files are not formatted at the moment

### LINKED ISSUE #171: init files are not formatted at the moment
The linters skip the init files, but they should be formatted too.

Moreover, `struphy format __init__.py` formats only `src/struphy/models/__init__.py`.
It removes the module docstring, which is inconvenient.

max-models:

Closes #171, #172.

Reformatted the init files and added a `struphy build-init-files` to rebuild them after adding new models or propagators.

In the near future, I think we can completely get rid of the `struphy format` command, and just use `ruff format` for formatting and `ruff check` for linting instead.

### DIFF
--- a/src/struphy/console/format.py
+++ b/src/struphy/console/format.py
@@ -1278,60 +1278,70 @@ def run_linters_on_files(linters, python_files, flags, verbose):
                     print(line, end="")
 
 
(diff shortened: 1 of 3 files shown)


spossann (approved):
Nice!

PR #287, opened 2026-07-23
Converted som kernels to return values instead of in-place mutation

### LINKED ISSUE #134: Convert all pyccelized kernels to functions
If possible, I think we should avoid updating input arrays in place in the pyccelized kernels. This would make the cupy/numpy interoperability more streamlined since we just need to convert to returned values back to cupy arrays, and we don't have to rely on possibly changed arrays in the arguments (which need to be converted from cupy to numpy before they are passed to a pyccelized kernel).

max-models:
This PR adds returns to some of the pyccelized kernels instead of updating (small) arrays in-place.

Related to #134, but some kernels don't really fit. So I just converted a few of the obvious kernels.

Why do this? Firstly, I think it makes the code a bit easier to follow, because it's not always obvious which arrays are mutated and which are not. Secondly, for the cupy/numpy switch, we still have to do something to convert the arrays to numpy before calling the kernel. Afterwards, some of the arrays may need to be converted back to cupy, but in principle this is only needed for the mutated arrays, which we don't have a clear way of specifying. If we instead always return one (or multiple) arrays, then it's always clear that the returned numpy array should be converted to cupy afterwards.

For some pyccelized kernels, like accumulation kernels, I don't think it makes sense to return new arrays, since they are too big, so only in-place mutation is realistic. Maybe we can update the `Pyccelkernel.__call__` method in way where we can specify which arrays should be mutated and which not. I'm not sure of what is best here.

### DIFF
--- a/src/struphy/feec/local_projectors_kernels.py
+++ b/src/struphy/feec/local_projectors_kernels.py
@@ -5,7 +5,7 @@ import struphy.kernel_arguments.local_projectors_args_kernels as local_projector
 from struphy.kernel_arguments.local_projectors_args_kernels import LocalProjectorsArguments
 
 
-def compute_shifts(IoH: "bool[:]", p: "int[:]", B_nbasis: "int[:]", shift: "int[:]"):
+def compute_shifts(IoH: "bool[:]", p: "int[:]", B_nbasis: "int[:]") -> "int[:]":
     """This function computes by how much we must shift the indices in case we loop over the evaluation points.
 
     Parameters
@@ -19,9 +19,13 @@ def compute_shifts(IoH: "bool[:]", p: "int[:]", B_nbasis: "int[:]", shift: "int[
     B_nbasis: 1d int array
         Array with the number of B-splines in each direction.
 
-    shifts : 1d int array
+    Returns
+    -------
+    shift : 1d int array
         array of 3 ints, each one denotes the amout by which we must shift the indices to loop around the quasi-points for each spatial direction.
     """
+    shift = zeros(3, dtype=int)
+
     for i, ioh in enumerate(IoH):
         # Histopolation
         if ioh:
@@ -32,6 +36,8 @@ def compute_shifts(IoH: "bool[:]", p: "int[:]", B_nbasis: "int[:]", shift: "int[
         # Interpolation
         else:
             shift[i] = -2 * B_nbasis[i]
+
+    return shift
 
 
 def get_local_problem_size(periodic: "bool[:]", p: "int[:]", IoH: "bool[:]"):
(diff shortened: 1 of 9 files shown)


src/struphy/feec/local_projectors_kernels.py
```diff
@@ -107,14 +115,21 @@ def get_dofs_local_1_form_ec_component_weighted(
...
             This int tell us whichone of the three components of the 1-form vector we are dealing with. It must be 0, 1 or 2.
+
+    Returns
+    -------
+    f_eval_aux : 3d float array
+        The evaluated degrees of freedom.
     """
     p = args_solve.degree[c]
 
+    f_eval_aux = zeros((n1, n2, n3), dtype=float)
```
spossann, line 131: It is possible though that we do not want to allocate a new array every time we call this function. Especially for large arrays, which should be allocated only once at the start. In that case one should pass `out` to the kernel. 

As a rule, we should pass `out` whenever the array has stencil vector size, 1d, 2d or 3d.

spossann (changes requested):
We do no want to allocate arrays too often, see my comment below. For stencil date size objects we want to pass `out`.

src/struphy/feec/local_projectors_kernels.py
max-models: I agree, I think I have to rethink this a bit. There will definitely be a lot of cases where we would like to pass an `out`.

Maybe we should add an `outputs: tuple` argument to the PyccelKernel class https://github.com/struphy-hub/struphy/blob/59a288f2a1326dc5b4910fe45c36d0936dce2263/src/struphy/utils/pyccel.py

In that way, we can do:

```python
interpolate = Pyccelkernel(
    some_interpolation_kernel,
    outputs=(5,),
)

interpolate(
    x,
    y,
    z,
    basis,
    coeffs,
    out,      # argument 5
)
```

And inside the `__call__` method of `Pyccelkernel`, we **only** convert the `outputs` arrays back to numpy arrays:

```python
def __call__(self, *args: Any, **kwargs: Any) -> Any:
    if self.use_cupy:
        # ...

        result = self._kernel(*args_np, **kwargs_np)

        # Only copy back mutated arrays
        for i in self._outputs:
            if isinstance(args[i], xp.ndarray):
                args[i][...] = xp.asarray(args_np[i])

        # ...
```

spossann: Yes something along these lines is needed 👍

max-models: Ok, I will try that!

max-models: I'm adding this functionality to `cunumpy` so that it can be used in both struphy and feectools https://github.com/max-models/cunumpy/pull/22

PR #302, opened 2026-07-29
300 add a poisson solver case to the profiling examples

### LINKED ISSUE #300: Add a Poisson solver case to the profiling examples
I only tested the profiling script on Pitagora, so if you want to run it on Raven/Viper, we probably have to make some small changes in the slurm parameters.

spossann:

Closes #300


max-models (dismissed):

max-models:
When you agree, then I'll merge this branch @spossann

max-models (approved):

max-models:
Here is an example of the strong scaling test: https://struphy-hub.github.io/profiling-data/figures/?case=20260730T094119Z-5e4d2dda-poisson_cube_strong_scaling#cases

Clearly we should increase the size of the  problem, but I think we can do that in a separate PR after we merged this one and #301

spossann:
I will now add some post-processing to the test case, to be sure that the result is correct.

max-models:
Do you want to add a postprocessing and verification step to this PR before merging? @spossann

spossann:
> Do you want to add a postprocessing and verification step to this PR before merging? @spossann

Ok, we can do it in a separate PR, let's merge. I have colloquium now, will be absent unitl 13h00.

max-models:
> > Do you want to add a postprocessing and verification step to this PR before merging? @spossann
> 
> Ok, we can do it in a separate PR, let's merge. I have colloquium now, will be absent unitl 13h00.

Ok!

max-models (approved):

Issue #290, opened 2026-07-24
Adapt code to ruff 0.16.0


max-models:
Since the [release of ruff 0.16.0](https://github.com/astral-sh/ruff/releases/tag/0.16.0), we have many errors with `ruff check`.

I think we should start by pinning ruff to 0.15.0, but later updating the code to also pass all the checks in 0.16.0.

Issue #18, opened 2025-10-23
Rewrite `dependencies.py`


max-models:
Currently, this file "works", but only if kernels are imorted in a very specific way ([which actually for a period was not supported by pyccel](https://github.com/pyccel/pyccel/issues/2353)).

There must be a better way of checking the dependencies.

max-models:
This should be easier now with pyccel make https://github.com/pyccel/pyccel/pull/2383

PR #306, opened 2026-07-30
Ran ruff check --fix with ruff 0.16.0

max-models:
Related to #290

This PR just contains automatic fixes by ruff 0.16.0, so there should not be any risks.

Next, we should also take care of the non-automatic fixes so we comply with all the ruff 0.16.0 rules.

Note that this formatting changes a few smart cases such as 

before:

```python
elif (Nbasis - p) < i and i < (Nbasis - 1):
```

after:
```python
elif (Nbasis - p) < i < (Nbasis - 1):
```


max-models:
Apparently it wasn't safe https://github.com/struphy-hub/struphy/actions/runs/30520367676/job/90799156512?pr=306

max-models:
The issue is that ruff changed the imports in the pyccelized kernels which changes the way our `dependency.py` file orders the compilation.

Therefore, in order to finish this PR, we first have to move to the build in makefile system in recent versions of pyccel and use that instead of our own makefile. See #18 for example.

max-models:
> The issue is that ruff changed the imports in the pyccelized kernels which changes the way our `dependency.py` file orders the compilation.
> 
> Therefore, in order to finish this PR, we first have to move to the build in makefile system in recent versions of pyccel and use that instead of our own makefile. See #18 for example.

Reverted to old import style in [73c6a01](https://github.com/struphy-hub/struphy/pull/306/commits/73c6a0199e6be699d5ee03d156dbcdb3e6d61665)

PR #313, opened 2026-08-03
310 add postprocessing to the poisson profiling case

### LINKED ISSUE #310: Add postprocessing to the Poisson profiling case
I will now add some post-processing to the test case, to be sure that the result is correct.

_Originally posted by @spossann in https://github.com/struphy-hub/struphy/issues/302#issuecomment-5140465320_

spossann:

Closes #310 

Assertion for the relative errors for solution and rhs have been added. If these assertions pass, figures and errors are stored in `sim.env.sim_folder`. Should we push them to the server for later inspection?


max-models (approved):
Looks good! These figures will be included in the `profiling-data` once https://github.com/struphy-hub/struphy/pull/315 is merged.

PR #317, opened 2026-08-03
Added memory estimates in struphy and feectools

### LINKED ISSUE #316: Allow to estimate memory usage before allocate
New method like `sim.estimate_mem()` that can be called **BEFORE** `sim.allocate()`.

For this, we should have an abstract method `Variable.estimate_mem()` for each variable. The screen output can show the mem usage of each variable + overhead from other parts (pproc? aux variables...)

max-models:
Corresponding PR in feectools: https://github.com/struphy-hub/feectools/pull/75

Example:

```bash
>>> from struphy import Simulation
>>> from struphy.models import Maxwell
>>> sim = Simulation(model=Maxwell())
>>> mem = sim.estimate_mem()
>>> mem = sim.estimate_mem(print_report=True)

ESTIMATED MEMORY USAGE (before allocate()):
  em_fields.e_field: 0.02 MB (local, rank 0)
  em_fields.b_field: 0.02 MB (local, rank 0)
  matrices.derivatives: 0.00 MB (local, rank 0)
  matrices.M0: 0.20 MB (local, rank 0)
  matrices.M1: 0.61 MB (local, rank 0)
  matrices.M2: 0.61 MB (local, rank 0)
  matrices.M3: 0.20 MB (local, rank 0)
  matrices.Mv: 0.61 MB (local, rank 0)
  TOTAL: 2.27 MB (local, rank 0), 2.27 MB (global, summed over 1 rank(s))
>>> print(mem)
{'em_fields.e_field': 22464, 'em_fields.b_field': 22464, 'matrices.derivatives': 0, 'matrices.M0': 202176, 'matrices.M1': 606528, 'matrices.M2': 606528, 'matrices.M3': 202176, 'matrices.Mv': 606528, 'total': 2268864}
```


src/struphy/pic/base.py
```diff
@@ -471,6 +481,27 @@ def n_rows(self):
...
+        :attr:`n_cols`, so it is valid whether or not the arrays were actually allocated
+        (see the ``dry_run`` argument of :meth:`__init__`)."""
+        float_size = 8  # dtype=float
+        bool_size = 1  # dtype=bool
+        n_rows = self.n_rows
+        n_cols = self.n_cols
+
+        nbytes = 0
+        nbytes += n_rows * n_cols * float_size  # markers
+        nbytes += n_rows * n_cols * float_size  # sorting_etas (mpi_sort_markers buffer)
```
spossann: There is an inefficiency uncovered: `sorting_etas` can be of shape `(self.markers.shape[0], 3)`, just as `is_on_proc_domain` - please adapt in the code and see if tests still pass.

spossann (commented):
That looks good so far. However, there is a big piece missing: the FEEC matrices, which are way bigger than the vectors! The best way to track them is I guess within `StencilMatrix.__init__()`, which could be augmented with a `dry_run` argument, returning before `self._data     = np.zeros(dims+diags, dtype=W.dtype)` is called.

src/struphy/pic/base.py
max-models: Yes, of course! This is the priority for today!

max-models:
In the workflows, we check the submodule diffs only after installing struphy.

However, in this PR, the struphy installation fails because it relies on the feectools version in the submodule, see https://github.com/struphy-hub/struphy/actions/runs/30913232722/job/92004850136?pr=317#step:9:986 for example.

It would be nice to change the order so we install feectools first, but I think submodule-diff currently depends on the install step because the current branch is fetched and checked out in struphy_in_container, what do you think could be a solution? @spossann 

```yaml
      - name: Install Struphy in Container
        uses: ./.github/actions/install/struphy_in_container

      - name: Get submodule diff
        uses: ./.github/actions/submodule-diff
        with:
          start-dir: /struphy_fortran_

      - name: Reinstall feectools from submodule
        if: env.SUBMOD_CHANGED == 'true'
        uses: ./.github/actions/install/feectools-submodule
        with:
          env-name: /struphy_fortran_/env_fortran_
```

src/struphy/pic/base.py
max-models: Done in [44242cf](https://github.com/struphy-hub/struphy/pull/317/commits/44242cf7fb104a7c567734fc7c6578d05e0c6872)

max-models:
> In the workflows, we check the submodule diffs only after installing struphy.
> 
> However, in this PR, the struphy installation fails because it relies on the feectools version in the submodule, see https://github.com/struphy-hub/struphy/actions/runs/30913232722/job/92004850136?pr=317#step:9:986 for example.
> 
> It would be nice to change the order so we install feectools first, but I think submodule-diff currently depends on the install step because the current branch is fetched and checked out in struphy_in_container, what do you think could be a solution? @spossann
> 
> ```yaml
>       - name: Install Struphy in Container
>         uses: ./.github/actions/install/struphy_in_container
> 
>       - name: Get submodule diff
>         uses: ./.github/actions/submodule-diff
>         with:
>           start-dir: /struphy_fortran_
> 
>       - name: Reinstall feectools from submodule
>         if: env.SUBMOD_CHANGED == 'true'
>         uses: ./.github/actions/install/feectools-submodule
>         with:
>           env-name: /struphy_fortran_/env_fortran_
> ```

Actually, the error doesn't come from the install, it comes from this line after installing:

```
STRUPHY_PATH=$($PYTHON -c "import struphy; print(struphy.__path__[0])")
```

spossann:
> > In the workflows, we check the submodule diffs only after installing struphy.
> > However, in this PR, the struphy installation fails because it relies on the feectools version in the submodule, see https://github.com/struphy-hub/struphy/actions/runs/30913232722/job/92004850136?pr=317#step:9:986 for example.
> > It would be nice to change the order so we install feectools first, but I think submodule-diff currently depends on the install step because the current branch is fetched and checked out in struphy_in_container, what do you think could be a solution? @spossann
> > ```yaml
> >       - name: Install Struphy in Container
> >         uses: ./.github/actions/install/struphy_in_container
> > 
> >       - name: Get submodule diff
> >         uses: ./.github/actions/submodule-diff
> >         with:
> >           start-dir: /struphy_fortran_
> > 
> >       - name: Reinstall feectools from submodule
> >         if: env.SUBMOD_CHANGED == 'true'
> >         uses: ./.github/actions/install/feectools-submodule
> >         with:
> >           env-name: /struphy_fortran_/env_fortran_
> > ```
> 
> Actually, the error doesn't come from the install, it comes from this line after installing:
> 
> ```
> STRUPHY_PATH=$($PYTHON -c "import struphy; print(struphy.__path__[0])")
> ```

This line could be replaced by sg like `pip show struphy | grep Location`.

max-models:
> > > In the workflows, we check the submodule diffs only after installing struphy.
> > > However, in this PR, the struphy installation fails because it relies on the feectools version in the submodule, see https://github.com/struphy-hub/struphy/actions/runs/30913232722/job/92004850136?pr=317#step:9:986 for example.
> > > It would be nice to change the order so we install feectools first, but I think submodule-diff currently depends on the install step because the current branch is fetched and checked out in struphy_in_container, what do you think could be a solution? @spossann
> > > ```yaml
> > >       - name: Install Struphy in Container
> > >         uses: ./.github/actions/install/struphy_in_container
> > > 
> > >       - name: Get submodule diff
> > >         uses: ./.github/actions/submodule-diff
> > >         with:
> > >           start-dir: /struphy_fortran_
> > > 
> > >       - name: Reinstall feectools from submodule
> > >         if: env.SUBMOD_CHANGED == 'true'
> > >         uses: ./.github/actions/install/feectools-submodule
> > >         with:
> > >           env-name: /struphy_fortran_/env_fortran_
> > > ```
> > 
> > 
> > Actually, the error doesn't come from the install, it comes from this line after installing:
> > ```
> > STRUPHY_PATH=$($PYTHON -c "import struphy; print(struphy.__path__[0])")
> > ```
> 
> This line could be replaced by sg like `pip show struphy | grep Location`.

I did a similar fix in [be1815c](https://github.com/struphy-hub/struphy/pull/317/commits/be1815cf72be72fc61c83421dbe4ad0fb6f7c6af) [f60b0cd](https://github.com/struphy-hub/struphy/pull/317/commits/f60b0cdeee1ffec8a94c2b9f69f04ec0081e11cf)

spossann (approved):
looks good!

PR #318, opened 2026-08-03
Nightly automatic check of dependency bounds

max-models:
This PR runs `utils/check_dependency_bounds.py` each night and opens a PR with the updated dependency bounds (from `utils/update_dependency_bounds.py`) if  any updates are available.


spossann (approved):
It makes a lot of sense to automate that - nice!

PR #320, opened 2026-08-03
Enforce that feectools submodule should be pointing to the last commit

max-models:
Adds a PR-time GitHub Actions workflow to enforce that the `feectools` git submodule in this repo is pinned to the latest commit on `struphy-hub/feectools`’s `devel-tiny` branch.

**Changes:**
- Introduces a new PR workflow that compares the `feectools` gitlink SHA in the superproject against the remote `devel-tiny` HEAD.
- Emits a failure summary with local commands to update the submodule when it’s out of date.


spossann (approved):
Yes that is what we need.

PR #322, opened 2026-08-03
Only allocate the full eval grid on rank 0

### LINKED ISSUE #319: Don't allocate the full eval grid on each rank in pproc

max-models:

Closes #319 

This PR rewrites the field reconstruction in the postprocessing one snapshot at a time.

**Changes:**

- One SplineFunction set is allocated once (`fields[species][var]`, not `fields[t][...]`) and refilled per step by the new `_load_femfields`. Files are opened once via [`ExitStack`](https://docs.python.org/3/library/contextlib.html#contextlib.ExitStack).
- Grids are now rank local, `_create_eval_grids` returns global grids plus per-rank slices from `derham.domain_array`, so each rank holds only `grids_log_loc`. The physical grid is built on rank 0 only.
- Gather replaces allreduce in `_collect_on_root`, there was no need for `MPI.SUM`.
- `_eval_femfields` rewritten for a single snapshot.
- Cleanup: `celldivide` is now `int | Sequence[int] = (1, 1, 1)`. If set to an int then is converted to `celldivide = (celldivide,) * 3`
- Added tests

The grid dump and vtk output is still handled by only rank 0.


src/struphy/post_processing/post_processing_tools.py
```diff
@@ -368,18 +373,71 @@ def process_fields(
...
+        single snapshot are held in memory, and each rank evaluates only those points of the
+        evaluation grid that lie in its own MPI domain. Arrays of the size of the global
+        evaluation grid therefore only ever exist on rank 0, where they are needed for output.
+
+        Parameters
+        ----------
+        step : int
+            Interval of saved time steps to post-process (1 = every step, 2 = every second step, ...).
+        celldivide : int
+            Grid refinement factor when evaluating FEM fields.
```
spossann: Maybe this should be a `Sequence`.

spossann (commented):
This looks great! I assume you are going to add a unit test?

src/struphy/post_processing/post_processing_tools.py
max-models: Done in [dd85c1c](https://github.com/struphy-hub/struphy/pull/322/commits/dd85c1c4d9ce082069c5010154c134c0b640cabb) and [1dfd780](https://github.com/struphy-hub/struphy/pull/322/commits/1dfd780d31dfa64b2ea45b11a37bfd87821ed5af)

spossann (approved):

PR #336, opened 2026-08-09
Added setters to all properties of the `Simulation` class

max-models:
New setters for model, name, description, params_path, time_opts, domain, equil, grid, derham_opts, plus new comm and start_time properties/setters.

### DIFF
--- a/src/struphy/simulation/sim.py
+++ b/src/struphy/simulation/sim.py
@@ -1732,6 +1732,7 @@ if __name__ == "__main__":
 
     @model.setter
     def model(self, value: StruphyModel):
+        assert isinstance(value, StruphyModel)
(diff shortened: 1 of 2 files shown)


src/struphy/simulation/sim.py
```diff
@@ -1776,21 +1730,44 @@ def model(self) -> StruphyModel:
         """StruphyModel object containing the PDE of the model."""
         return self._model
 
+    @model.setter
+    def model(self, value: StruphyModel):
+        assert hasattr(value, "propagators"), "Attribute 'self.propagators' must be set in model __init__!"
```
spossann, line 1736:
```suggestion
        assert isinstance(value, StruphyModel)
        assert hasattr(value, "propagators"), "Attribute 'self.propagators' must be set in model __init__!"
```

spossann (dismissed):
Nice!

PR #337, opened 2026-08-10
Compressible two fluid model

szekedidavid:


src/struphy/feec/linear_operators.py
```diff
@@ -491,14 +509,28 @@ def dot(self, v, out=None):
...
+            if out is not None:
+                assert isinstance(out, Vector)
+                assert out.space == self._codomain
+                v.copy(out=out)
+            elif self._cross_space:
+                out = self._codomain.zeros()
+                v.copy(out=out)
+            else:
+                out = v.copy()
+            return out
```
spossann: This needs to be removed. I will try to understand why `H1` needs a different treatment here.

PR #339, opened 2026-08-11
Merged PIC/feec openMP command and added back a few openMP pragmas

max-models:


src/struphy/utils/test_mpi_launch.py
max-models, line 1: can be removed, it's from an old PR

spossann (approved):

PR #347, opened 2026-08-17
Fixed the `xp.all(axis=1)` bottleneck

max-models:
Reducing over a size-3 trailing axis across millions of rows is slow with both numpy and cupy. The `np.all(axis=1)` can't vectorize a 3-wide reduction, while the 3 column slices is a large and easily vectorizable operation.


spossann (approved):
Nice find!

PR #354, opened 2026-08-20
Update profiling setup

max-models:
Updated scope-profiler to 0.3.3. There are some big improvements in the postprocessing in this version. Especially for GPU profiling and also line-by-line profiling for the regions we already defined using the tool without having to use line_profiler directly.

For example, please try running this param file:

```python
from struphy import ProfilingOptions, Simulation
from struphy.models import Poisson

model = Poisson()

profiling_opts = ProfilingOptions(use_line_profiler=True)

sim = Simulation(
    model=model,
    profiling_opts=profiling_opts,
)

sim.run(profiling_activated=True)
```

Then inspect the profiling file in the terminal with

```
scope-profiler tui sim_1/profiling_data.h5
```

image

With `scope-profiler > 0.3.0`, there is a new optional syntax for the profiling manager. Instead of writing `ProfilingManager.setup(...)` and then `ProfilingManager.finalize()`, it is now possible to define the start and stop of the profiling manager work using contexts, just as is done with the profiling regions.

For example (from the [scope-profiler readme]()

```python
from scope_profiler import ProfileManager

with ProfileManager.session():
    @ProfileManager.profile("main")
    def main():
        with ProfileManager.profile_region("work"):
            sum(range(100))  # replace with the code you want to measure

    main()
# writes profiling_data.h5 and prints a summary
```

For us, this means we can use this syntax in `sim.run()`, so we don't need the `_setup_profiling()` method. Everything in `sim.run()` is just indented by one tab after `with ProfileManager.session(...):`, but in the diff it looks like there is a much bigger diff... I promise there isn't though.

Changes in the code:

1. Added `functions=[...]` to `ProfileManager.profile_region` calls which are profiling a very small scope, most likely just a function. When line profiling is turned on, this means that the lines inside that function are profiled, not just the line for the function call itself.

For example:

```python
with ProfileManager.profile_region(self._solve_region, functions=[self._A_inv.solve]):
            _u = self._A_inv.solve(rhs, out=self._u_new)
```

2. Added a `ProfilingOptions` class to specify profiling option in the parameter files, this class is automatically added to the scripts with `struphy params MODEL`. This class basically mirrors the parameters used in `scope-profiler`
3. Deprecated `profiling_activated ` from the `EnvironmentOptions` class and moved it to `sim.run(profiling_activated=profiling_activated)` instead
4. Added `tutorials/tutorial_scope_profiling.ipynb`
5. Updated docs
6. Updated dependencies for gvec, vtk, feectools. I also set lower bounds for feectools and scope-profiler.


spossann:
That looks great - looking forward to testing it!

PR #365, opened 2026-09-11
Create a new simulation file directly in test_pproc.py

max-models:
Should fix the macos scheduled error


spossann:
Another solution is implemented here: https://github.com/struphy-hub/struphy/pull/363/changes#diff-26cec124f37976a86b1b973618ff35f82ec1a040af81ff7643db1c58eddd883aR16

spossann:
Has been fixed in https://github.com/struphy-hub/struphy/pull/363

PR #370, opened 2026-09-12
Postprocessing

max-models:
See updated guide here: https://github.com/struphy-hub/struphy/blob/postprocessing/doc/markdown/output-api.md

# Working with simulation output

`Output` is a lightweight handle to one completed Struphy run. It reads saved metadata when it
is created, but does not load field or particle data until requested.

```python
from struphy import Output

out = Output("path/to/run")
```

When working directly after a simulation, `Simulation.run()` returns the same kind of object:

```python
out = sim.run()
```

## Discover available data

Use `keys()` to list the names accepted by `evaluate()`, or `info()` for the same names with
short descriptions.

```python
print(out.info())

for key in out.keys():
    print(key)
```

`keys()` and `info()` do not load product arrays. If post-processed products do not exist yet,
they materialize them using default post-processing options. Call `pproc()` first when those
options matter.

## Materialize post-processing products

Use `pproc()` explicitly to choose how fields and particle diagnostics are generated.

```python
out.pproc(
    physical=True,       # also create physical field components and coordinates
    step=1,              # use every saved time step
    celldivide=1,
)
```

Matching existing products are reused. `evaluate()` also calls `pproc()` automatically for a
missing non-scalar product when running serially.

## Evaluate data

`evaluate()` returns an ordinary `xarray.DataArray`. Use xarray for selections, arithmetic,
reductions, and interoperability with other scientific Python packages.

```python
rho = out.evaluate("diagnostics/rho_xyz")
phi = out.evaluate("phi_integral")

# Last saved time and one vector component, selected by integer position
electric_field = out.evaluate("em_fields/E", isel={"t": -1, "component": 2})

# Select the nearest logical-coordinate plane
midplane = out.evaluate(
    "diagnostics/rho_xyz",
    sel={"e3": 0.5},
    method="nearest",
    drop=True,
)

# Select a time range
history = out.evaluate("phi_integral", isel={"t": slice(100, None)})
```

`isel` uses integer positions and `sel` uses named dimension-coordinate values. Selections are
applied in that order. Physical `X`, `Y`, and `Z` coordinates describe the mapped logical grid;
evaluating at an arbitrary physical point requires interpolation or an inverse-coordinate map.

To return only values, without xarray coordinates and attributes, use `as_numpy=True`.

```python
rho_values = out.evaluate("diagnostics/rho_xyz", isel={"t": -1}, as_numpy=True)
```

`out.xarray` provides the complete lazy xarray `DataTree` when access to the grouped product
store is useful. Prefer `evaluate(key)` for normal single-product work.

## Plot data

Struphy-aware plotting is performed by `Output`, not by modifying xarray arrays. Rendering
methods return `(fig, ax)` (or `(fig, axes)` for panels), so normal Matplotlib controls display,
saving, and further customization.

```python
from matplotlib import pyplot as plt

fig, ax = out.timeseries("phi_integral", fit=(0.0, None), fit_amplitude=True)
ax.set_title("Potential growth")

fig, ax = out.viewer(
    "diagnostics/rho_xyz",
    x="e1",
    y="e2",
    coords="physical",
    plane="RZ",
)

fig, axes = out.panels("kinetic_ions/e1_v1_density/f", x="e1", y="v1")
fig, ax = out.trajectories("kinetic_ions", max_markers=1000)

plt.show()
```

Plotting methods also accept a derived `DataArray` instead of a saved-product name.

```python
rho_last = out.evaluate("diagnostics/rho_xyz", isel={"t": -1})
fig, ax = out.slice(rho_last, x="e1", y="e2", coords="physical", plane="RZ")
```

The available product plotting methods are `timeseries`, `slice`, `panels`, `viewer`,
`animation`, and `trajectories`. `view` creates a reusable view configuration, while `frames`
writes PNG files and returns their paths. Whole-run plots are `plot_scalars` and `equilibrium`.

## MPI post-processing

`Output` always uses `MPI.COMM_WORLD`; no communicator is passed to its constructor.

For serial post-processing under MPI, call `pproc()` on every rank. Rank 0 does the work and the
other ranks wait at the synchronization barrier.

```python
out.pproc(physical=True)
```

For parallel post-processing, also call it on every rank and pass `parallel=True`. The current
world communicator must have the same number of ranks as the run that wrote the raw output.

```python
out.pproc(parallel=True, physical=True)
```

Automatic materialization through `evaluate()` is intentionally disabled when more than one MPI
rank is active. Call `pproc()` explicitly first in that case.


spossann:
Does this solve the doc build error?? https://github.com/struphy-hub/struphy/actions/runs/35831426449/job/107084726816

src/struphy/diagnostics/diagn_tools.py
```diff
@@ -16,13 +26,66 @@
 logger = logging.getLogger("struphy")
 
 
+def _legacy_plot(replacement):
```
spossann: I feel like we should remove the whole legacy business here. It will not be hard to replace the few instances where this is used with the new code.

max-models:
> Does this solve the doc build error?? https://github.com/struphy-hub/struphy/actions/runs/35831426449/job/107084726816

No, only the model test error https://github.com/struphy-hub/struphy/actions/runs/35839448887/job/107110743967

src/struphy/diagnostics/tests/__init__.py
spossann: Since this is a new test folder, it must be included in a shard manually:

- https://github.com/struphy-hub/struphy/blob/f2edd752b0805f98ffafb3ee15e1b9a761faadde/.github/workflows/reusable-unit-testing.yml#L39
- https://github.com/struphy-hub/struphy/blob/f2edd752b0805f98ffafb3ee15e1b9a761faadde/.github/workflows/reusable-scheduled.yml#L107

src/struphy/diagnostics/plotting.py
max-models: Maybe move all the plotting to a new repo `struphy-plots`

src/struphy/feec/mass.py
max-models: This will be merged into devel in https://github.com/struphy-hub/struphy/pull/384

spossann:
Some thougjts on `Output.evaluate()`:

- It should take a string on the form `species/variable` as a mandatory arg. 
-  **It should always return an Xarray**
- We already discussed the `FEECVariable` case. The Xarray should in addition always contain the X, Y and Z coords for plotting in physical space. These are derived from the eta inputs by evaluating the domain. The default etas should be the mid points of the sim grid.
- In the `PIC/SPHVariable` case, there could be binning-, kernel-density- and/or orbit data. By default, the returned Xarray should be the first binning plot it finds, at all time steps, then kd plot, then orbits if none of the others is found. As arguments one should be able to pass `t` as for feec, or a specific data set of the above. There shouldd be an `info` command that allows you to see the possible data to evaluate.

PR #385, opened 2026-09-20
bugfix: Fix nested scalar sums in Scalars.update()

max-models:
## Summary

Fix stale values in nested scalar sums during repeated `Scalars.update()` calls.

## Problem

Scalar expressions with three or more terms are nested:

```python
a + b + c
# SumOfScalars(SumOfScalars(a, b), c)
```

`Scalars.update()` previously marked only its top-level entries as outdated after an update. The inner `SumOfScalars(a, b)` is not a top-level entry, so it remained marked up to date.

On the next update, the outer sum called `inner_sum.update()`, which returned without recalculating. The total therefore combined a frozen value of `a + b` with the current value of `c`.

This affected derived quantities such as total energy, producing apparent energy drift even when the underlying propagators conserved their individual contributions.

## Fix

Recursively mark every scalar reachable through a sum as outdated after each container update:

```python
def _mark_outdated(scalar):
    scalar.uptodate = False
    for variable in scalar.variables:
        if isinstance(variable, Scalar):
            _mark_outdated(variable)
```

This ensures nested sums and their leaf scalars are recalculated at the next output/update time.

## Extra

This PR also fixes a second stale-value bug in the same file. `KineticEnergyPIC` and `KineticEnergySPH` cached `particles.velocities` and `particles.weights` on their first update. Both are fancy-indexed (`self.markers[self.valid_mks, self.index["vel"]]`), so they return copies, not views: the cached arrays hold the state of the first call and the reported kinetic energy never changes again. Verified on a short weak-landau-damping run, where `kinetic_energy` was bitwise constant while `electric_energy` fell by 99.8%. Affects every PIC and SPH model (`VlasovAmpereOneSpecies`, `VlasovMaxwellOneSpecies`, `Vlasov`, the SPH models and the hybrids) and with it every `total_energy` that includes a kinetic channel. Fixed by reading the marker arrays at each update, `valid_mks` needs no invalidation hook, since `_update_valid_mks()` already refreshes it in place.

## Validation

Added a regression test for `a + b + c`:

1. Update scalar values and calculate the total.
2. Change all three values.
3. Update again and confirm the total reflects all new values.

The test fails before this change and passes after it.


spossann:
The problem with `Scalars` was also recognized in https://github.com/struphy-hub/struphy/pull/276

spossann (approved):

